PR #19, opened 2015-01-25
Fixing layout issues and adding eslint.json for easier updates

ilyavolodin:
Removed jumbotron from demo page. Added height to the editor. Added eslint.json file and referenced it from the code. I hope that works better. I still have no idea how to test Jekyll pages, so more or less coding blind here.


nzakas:
Haha I know, it's such a pain. Let's see how it goes.

ilyavolodin:
Height still isn't working. Annoying. I guess it needs to be set through style, not attributes. I'll create another pull request soon.

PR #28, opened 2015-03-07
Adding configuration to eslint demo (fixes #26)

### LINKED ISSUE #26: Add rule configuration option to the demo
ESLint has a ton of different rules, I think it would be good to add ability for users to play with different rules on the demo page. For that, we need to expose a way to configure them. It would be even cooler, if we would allow to use plugins, but I don't think that's possible with the current setup.

ilyavolodin:
(fixes #26) Adding configuration to the site's demo page. The code is pretty ugly, but I think it should work. The only thing I'm not sure and can't verify is if popovers are going to work correctly.


nzakas:
Guess we will see :)

ilyavolodin:
Well, that didn't work... Working on fixing it now.

PR #34, opened 2015-03-20
Change documentation structure (fixes #32)

### LINKED ISSUE #32: Move and restructure documentation
As mentioned in #31 we should move some of the documentation out of ESLint repository and into this one. I suggest we not only move it, we also restructure some of it. I propose removing "Docs" link on the website all together and then:
* Moving "Architecture" under "Developer Guide".
* Renaming "Configuring" to "Configuration" and making it a top level menu, and then moving "Command Line" under it.
* Moving "Integrations" somewhere... Not sure where, maybe under "Configuration" or "About"?

After the move, the only folders that should be left in ESLint repository are: "Rules", "Developer Guide" and  "Configuring". Everything else should be under this repository. Ideally, I would also move out everything from under "Developer Guide" here, except for "Working with Rules" and "Working with Plugins", but I think that could be done a bit later, as a separate pull request.

How does that sounds?

ilyavolodin:
Fixes #32 
**Please do not merge this until just before you are ready to do new release for ESLint.**
Otherwise some of the documentation is going to be broken. This moves things around a bit. And also changes navigation menu for the site. Now navigation items include "User guide", "Developer guide", "Blog", "Demo" and "About". User and Developer guides are now dropdowns for ease of navigation. I also changed site name to "ESLint: Pluggable JavaScript linter".


nzakas:
How about doing this in steps instead of all at once?

We could:

1. Change the title
2. Duplicate content to it's new location
3. Change existing links to point to be locations
4. Update navigation
5. Remove content from old location

We can do 1 and 2, maybe 3 right now, add hold off on the rest until a new release?

ilyavolodin:
I wonder why did github stopped notifying me about comments in this repository...
I think there's only one file that is currently missing. I can just add that file manually. All the rest of the links will work just fine. We just need to make sure that https://github.com/eslint/eslint/pull/2116 is merged in before the next release. Would that work?

nzakas:
I'm just worried about changes that might happen to these pages before the next release. Keeping the docs in limbo until the next release seems like an invitation for trouble.

ilyavolodin:
Sorry, wasn't clear. If I add back one missing documentation file ("command-line-interface") then everything can be merged now, and everything will work just fine with new navigation and documentation changes. When the next release rolls around, it will automatically updated dynamic parts of the documentation if https://github.com/eslint/eslint/pull/2116 will get merged after modifications.

docs/command-line-interface/index.md
```diff
@@ -1,231 +0,0 @@
----
```
nzakas: I think this probably needs to live in the eslint repo, since if we make changes to the CLI, we'll want this page updated with the PR to reflect that change.

ilyavolodin: It does. It was a delete action for the old generated version of it.

ilyavolodin:
Ok, I updated both this pull request as well as the one in eslint. This pull request can be merged now. It will not break the site anymore, just change navigation and update documentation structure. Once pull request in the main repository merges in, following release documentation will be automatically update on the site in the correct spots.

nzakas:
Awesome, thanks.

PR #45, opened 2015-03-28
Adding support redirects for old urls that are still used

ilyavolodin:
Google webmanster tools are reporting a bunch of broken links from other sites. Those are caused by move of the documentation. I added support for 301s for some of the old pages that people linked to. Unfortunately, Jekyll doesn't make it easy to do. Creating those new files just to add `redirect_to` feels dirty, but I don't think there's any other way.

I'm not going to merge this one myself. @nzakas see if you OK with this and merge it if you are.


nzakas:
Fine by me.

PR #48, opened 2015-03-30
Adding GitHub Link to header

JedWatson:
Not sure if it's just me but I always go looking for the link to the GitHub project in the header, then remember to scroll down to the footer to find it.

The GitHub project itself is particularly valuable when I'm trying to work out whether a certain rule is behaving as expected, is on the roadmap, etc. Would be nice to give it some more visibility.


nzakas:
Thanks, but we prefer it in the footer for now.

PR #49, opened 2015-03-30
Removed duplicate container

montogeek:
_layouts/default.html already defined a container class. Each container class add 15px of padding making the website looks odd.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3,7 +3,6 @@ title: ESLint - Pluggable JavaScript linter
 layout: default
 redirect_from: "/docs/"
 ---
-<div class="container">
 <div class="col-sm-7 col-md-12">
 
           <div class="row">
@@ -58,4 +57,3 @@ redirect_from: "/docs/"
               </div>
           </div>
     </div> <!-- /.col-md-12 -->
-</div>


nzakas:
Can you explain what looks odd?

montogeek:
screen shot 2015-03-30 at 15 27 05

The second `.container` add an extra padding that is unnecessary for the page.

Oh, in fact, it should be renamed to `.row`, following the BS grid.

ilyavolodin:
This is correct. Bootstrap container adds 15px padding on each side. Bootstrap row adds -15px padding left and right. When you have two containers nested you get 30px padding, so second container should be changed to row.

montogeek:
I noticed something else, for medium devices (width ≥ 992px), the page is resized to use the full space available
![Uploading Screen Shot 2015-03-30 at 15.32.12.png . . .]

montogeek:
I noticed something else, for medium devices (width ≥ 992px), the page is resized to use the full space available.

Changing it to `col-md-12` fixes it.

screen shot 2015-03-30 at 15 32 12

nzakas:
Do you want to make that change plus the change @ilyavolodin mentioned?

ilyavolodin:
Thanks for pointing out those issue @montogeek I just merged a pull request that fixed them and a few other things. Closing this one.

PR #58, opened 2015-04-08
Added more eslint integrations.

Globegitter:
Since I didn't see the integration.md in the official eslint repo I hope this PR is in the right place.

Having used 2 of these 3 plugins myself and planning to use the 3rd soon. So thought it would be useful for others.


ilyavolodin:
LGTM

PR #60, opened 2015-04-12
make urls links

also:
The URLs at the top of http://eslint.org/docs/developer-guide/source-code aren't being turned into links.


ilyavolodin:
Looks good. Thanks.

PR #68, opened 2015-05-24
Include anchor links on headings.

### LINKED ISSUE #67: Add Anchor links to sections in the docs
It would be nice to link [directly](http://ux.stackexchange.com/questions/36304/use-of-mouse-over-paragraph-marker-in-headlines-for-permalink) to various sections in the docs and developer guide via anchor links.

I happen to maintain [a tool](http://bryanbraun.github.io/anchorjs/) that helps with that and would be happy to help set it up.

bryanbraun:
Closes #67.


nzakas:
@ilyavolodin are you good with this?

ilyavolodin:
Bah, sorry, I was going to verify it and merge it if everything checks out. I'll try to do that tomorrow (left my mac at work).

ilyavolodin:
Sorry for the delay. Couldn't find the time to test this out earlier. Unfortunately, it doesn't seem to work as described. I checked a few rules, and a lot of our headers include special characters (mostly quotes). While the anchor link and icon show up on mouse-over for those, they don't link to anything. I checked just a few random rule pages, but they had a lot of those broken anchor links. (strict and dot-location rules, for example).

bryanbraun:
Interesting. It's designed to strip out special characters... maybe it's being too aggressive? I'll have a look.

bryanbraun:
Looks like you found a bug in the library! It's leaving the quote marks in the IDs, which breaks the html when the id attribute is updated. I'll apply a fix to strip out any single/double quotes and update the PR.

bryanbraun:
I take that back. The bug was actually in Redcarpet, and those IDs are currently broken in the Production Eslint site:

image

Redcarpet [has fixed the issue](https://github.com/vmg/redcarpet/pull/428), but it hasn't made it into a tagged gem release yet, so it isn't available in the Jekyll project. Even if it got tagged, Jekyll is a few versions behind the latest Redcarpet version, and once they get it updated you'd likely have to upgrade the Eslint site to a Jekyll 3.x, in order to get the fix (most new Jekyll fixes are going into the 3.x branch these days). 

Since you won't be able to anchor to the titles in their current state, I think it's better to just remove Redcarpet's generation of TOC IDs, and let AnchorJS do it (which it does by default if no ID is present). As a side benefit, AnchorJS already handles other situations that Redcarpet hasn't addressed [like duplicate IDs](https://github.com/vmg/redcarpet/issues/307), for example.

I'll push up a commit that does this... let me know what you think.

_includes/footer.html
```diff
@@ -14,5 +14,7 @@
 
   <script src="/js/vendor/jquery-1.10.2.min.js"></script>
   <script src="/js/vendor/bootstrap.min.js"></script>
+  <script src="/js/vendor/anchor.min.js"></script>
+  <script src="/js/app/main.js"></script>
```
ilyavolodin: Could you inline main.js here? The size of the header and overhead of the tcp-ip handshake is probably significantly larger in this case then the size of the file itself:-)

ilyavolodin:
Just one comment about removing main.js file. Everything else looks good. Is there a way to disable this on the homepage? It's pretty useless on the homepage, and in some cases just confusing. If not, it's not a huge deal, but would be nice.

bryanbraun:
Good suggestions. I've now targeted them only to appear on docs and blog pages (the demo page had a couple confusing ones as well).

ilyavolodin:
Looks good to me. Thanks for bearing with me:-)

bryanbraun:
:+1:

PR #75, opened 2015-06-12
Add saving of rules configuration on the demo page

EvgenyOrekhov:
I think this would make the demo page much more useful.


ilyavolodin:
Thanks. I'll merge it for now, but I do have plans to re-write the whole demo at some point, and add saving of the configuration into a file.

PR #76, opened 2015-06-12
Fix demo page performance issue

EvgenyOrekhov:
The "Ok" button had 164 identical click handlers attached to it, and they all were called on every click on the button.
Moved the click handler out of the for loop.

### DIFF
--- a/js/app/demo.js
+++ b/js/app/demo.js
@@ -146,27 +146,27 @@ require([
             if (i === limit) {
                 i = 0;
             }
-            $('#configuration .btn').click(function() {
-                var environments = {}, rules = {};
-                $('.environments input').each(function() {
-                    var name = $(this).attr('id');
-                    var value = $(this).is(':checked');
-                    environments[name] = value;
-                });
-                $('.rules input').each(function() {
-                    var name = $(this).attr('id');
-                    var value = $(this).is(':checked') ? 2 : 0;
-                    rules[name] = value;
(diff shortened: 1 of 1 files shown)


ilyavolodin:
Sorry, I might be blind, but I'm not seeing any difference between the code that's there and the code you submitted, other then line-breaks.

btmills:
@ilyavolodin the code was moved after the closing curly brace

EvgenyOrekhov:
@ilyavolodin, @btmills is right, I placed the code after the body of the `for (var rule in rules)` loop (after the curly brace on line 169/149 https://github.com/eslint/eslint.github.io/pull/76/files?diff=unified#diff-d494bd0caf730ff97f134507d6d3965cR149 ).

nzakas:
Was there a measurable performance problem that this addresses? Was it slow for you?

EvgenyOrekhov:
@nzakas yes, it executed 164 times slower before the fix. It took about 3 seconds, now it takes 16ms.

ilyavolodin:
Ahh.. I see now. Sorry, git compare made it pretty hard to notice. Yeah, looks like a valid fix. Having 164 handlers on a single button is not a good idea:-)

nzakas:
@EvgenyOrekhov thanks. In the future, please be sure to mention such details when reporting perf problems, it helps a lot.

PR #78, opened 2015-06-25
New: Espree Demo (fixes #77)

### LINKED ISSUE #77: Add Espree Parser
Let's do an updated version of this using espree:
http://esprima.org/demo/parse.html

We could borrow the style used here:
https://babeljs.io/repl/

xjamundx:
screen shot 2015-08-13 at 9 33 39 pm
screen shot 2015-08-13 at 9 33 31 pm

screen shot 2015-08-13 at 9 35 54 pm

- ES6 on by default
- Basic error support
- Not yet exposed `/parser` URL


nzakas:
Looks pretty! Maybe get that centered?

xjamundx:
Yes and eventually some kind of syntax highlighting

ilyavolodin:
Was just about to mention syntax highlighting. Do you want to just wire in orion like we did in the demo? Also, do we want to expose espree switches too?

xjamundx:
Yeah we def nef need all of the options. I think I can programmatically grab them and then generate the switches.

Also anyone care that I'm using webpack?

ilyavolodin:
Webpack, browserify - doesn't matter to me. They do more or less the same thing. I'm wondering if we should combine two demos into one. Type your code, get it validated by eslint, then switch tab on the right side and see AST

xjamundx:
@ilyavolodin I like the idea of a single demo with an option to switch between parser/linter

ilyavolodin:
@xjamundx Sounds good to me. Just put tabs on top of the area we have errors displayed in the current demo. One tab for errors, one tab for AST. Should be pretty straight-forward, I think. But we will need to add ecmaFeature switches below the editor.

IanVS:
If this will be doing linting as well and there are no default rules after 1.0, what would be the plan for configuring rules?  The easiest way may be to allow an `.eslintrc` file to be uploaded and used.  Or specify a sharable config to extend from?

nzakas:
We will use `eslint.json` as the source of rules enabled by default.

xjamundx:
I'm going to work on this again today/tomorrow. Really excited about getting this out there. Should hopefully help the debugging / AST explaining process as well :)

xjamundx:
Here are some new updates. Any use in shipping a minimal version until I get around to adding toggleable options?

xjamundx:
I'll vertically align those boxes in the morning... :)

ilyavolodin:
Yes, let's add a minimal version for now. It's better then nothing. Do you want to add it to the menu somewhere? Also, you might want to update Espree release script to upload a new version of parser.

xjamundx:
Give me an hour this morning to make a few adjustments and try to add a link somewhere...not sure how to do the build script. Let's add that late.r.

xjamundx:
screen shot 2015-08-14 at 10 00 26 am
screen shot 2015-08-14 at 10 00 20 am
screen shot 2015-08-14 at 10 00 05 am

BYK:
:+1: :astonished:

nzakas:
Looks nice!

xjamundx:
Also uses flebox, because we're for devs and on github

package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "eslint.github.io",
+  "version": "1.0.0",
+  "description": "This contains the code running on eslint.org.",
+  "main": "index.js",
+  "directories": {
+    "doc": "docs"
+  },
+  "scripts": {
+    "build": "webpack --module-bind json js/app/parser/index.js js/app/parser/index.built.js",
```
ilyavolodin: Sorry, I'm not familiar with WebPack, can you explain what this is suppose to do?

package.json
```diff
@@ -0,0 +1,32 @@
...
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/xjamundx/eslint.github.io.git"
```
ilyavolodin: This is not right.

xjamundx: so glad you checked. haha, no, that is not right

package.json
xjamundx: `npm run build` will output the built parser JS

ilyavolodin: Ahh.. ok, got it. I think this should be moved into Espree repository at some point (similar to browserify task in ESLint core), but I think we can merge it like this for now

nzakas: Why is this necessary? ESLint already bundles Espree with it. Can't you just `require('espree')` already?

ilyavolodin: Does browserified version of ESLint exposes Espree in some way that can be consumed?

nzakas: As I said, I believe you can just do `require('espree')`. Here's how the file is created: https://github.com/eslint/eslint/blob/master/Makefile.js#L561

xjamundx: Yeah I'm just calling `require('espree')` we don't need a special version

ilyavolodin: In that case, do you want to remove index.build.js and webpack script and just use eslint.js instead?

xjamundx: not so simple, but let me give soemthing a try

xjamundx:
```
~/dev/eslint.github.io (espree) $ browserify -x espree js/app/parser/index.js > js/app/parser/index.built.js 
Error: Cannot find module 'espree/package.json' from '/Users/jamuferguson/dev/eslint.github.io/js/app/parser'
    at /usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:46:17
    at process (/usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:173:43)
    at ondir (/usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:188:17)
    at load (/usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:69:43)
    at onex (/usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:92:31)
    at /usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:22:47
    at FSReqWrap.oncomplete (fs.js:95:15)
```

This method sort-of works. Requires some modifications.

xjamundx:
This should work now without regular re-builds, automatically taking in the latest from `eslint.js`. `npm run build` is there in case changes need to be made to the parser demo itself, but not for general changes to espree.

xjamundx:
works in safari/firefox/chrome at least

ilyavolodin:
Let's go with this.

gyandeeps:
Awesome... This would be very helpful and now I dont have to go to esprima website to see the ast.
:tada:

PR #84, opened 2015-07-03
Docs: add eslint-plugin-react to integrations

hzoo:


ilyavolodin:
Wow, it wasn't there already? Always though it was in the list. Thanks

PR #86, opened 2015-07-15
Adding users to homepage (fixes #33)

### LINKED ISSUE #33: Add users page
I really like what Babel did with this page: http://babeljs.io/users/

I'd like to have something similar for the site.

ilyavolodin:
Added users section to the homepage with 5 logos. Unfortunately, this script will execute on every page, but should only insert logos on the homepage. Couldn't find another way to do this with Jekyll.


nzakas:
You should be able to add this to the homepage front matter:

```yaml
homepage: true
```

Then in the template:

```
{% if page.homepage %}
Your script
{% endif %}
```

ilyavolodin:
Ahh.. Nice, I didn't realize that. Let me try it locally and submit a change if it works.

ilyavolodin:
Updated. Nice trick!

nzakas:
Lgtm. Merge at will.

PR #102, opened 2015-07-17
Fix: Logo spacing and sizing

gyandeeps:


gyandeeps:
It will improve spacing and now its going to look like this:

capture

ilyavolodin:
Yeah, that's looks better

PR #119, opened 2015-08-14
Fix: Add rule to error list (fixes #85)

### LINKED ISSUE #85: Demo: Should also include rule name in the error messages
I think it would be a good idea to include rule names inside the error messages. It will be consistent based on how eslint works and also it helps to find out which rule is throwing errors.

xjamundx:
screen shot 2015-08-13 at 9 01 23 pm

- Add rule name
- Add link to docs


ilyavolodin:
Cool. Thanks

PR #122, opened 2015-08-18
Fix: Missing Environments in Demo

xjamundx:
Hard-coding the list. Appears to work ok.


ilyavolodin:
We just need to make sure we update the list whenever we add new environments.

PR #127, opened 2015-08-28
Fix: Scroll long code instead of wrapping (fixes #126)

### LINKED ISSUE #126: Long lines of code in <pre> blocks wrap to next line
On smaller screens, or for longer lines of code, the current behavior is to wrap to the next line.  This reduces readability, and will become more of an issue when https://github.com/eslint/eslint/commit/0873e01b7240139ceac731e0b6232bf3f94fedfb hits the site, as there are some long error comments.  For example:

array-bracket-spacing_wrap

I propose instead to allow scrollbars like:

array-bracket-spacing_scroll

Note: @ilyavolodin has also suggested to only display the number of expected issues if there are more than one, but that can be discussed in a separate issue.  Regardless, small screens and long error messages will continue to cause problems unless this issue is addressed.

IanVS:
This unfortunately can't go into bootstrap.less, because it is overridden in `main.936b04d2.css`.  The other option is to edit that file (main), which is possible, but ugly.

### DIFF
--- a/styles/overrides.css
+++ b/styles/overrides.css
@@ -1,3 +1,13 @@
 .btn-getting-started {
   margin-top: 30px;
 }
+
+/* Prevent long code from wrapping, scroll instead */
+code[data-lang] {
+  white-space: pre;
+}
+
+main.doc .highlight pre {
+  overflow-x: auto;
+  word-wrap: normal;
+}


BYK:
>This unfortunately can't go into bootstrap.less, because it is overridden in main.936b04d2.css.

May be good to put this as a comment in the code ;)

IanVS:
I can add something to the top of `overrides.css`, as to what its purpose is.  If you weren't aware, the main css file we have is minified, so it's tough to work with.  I prefer adding overrides as I've done here, but would like to get the opinion of others on the team who have been working with it longer than I have.

ilyavolodin:
In all honesty, we should recreate bootstrap theme and check it into the repository. Designer who was doing the site compiled it locally and only uploaded compiled file, so now we can't really modify anything in original bootstrap.

IanVS:
I think @nzakas was talking at some point about an eventual re-design.  If that's the case, I'm not sure it would be worth re-creating the existing design at this point.  Seems like a lot of effort for not a whole lot of benefit.

As it stands, we can modify the design we have, but only by overriding as I've done here.  The minor downside I see is an additional css file.

ilyavolodin:
That "eventual" re-design might be far off into the future:-) But I do agree. I'm fine with overrides for now.

PR #141, opened 2015-09-11
Add Algolia logo

pixelastic:
This adds the Algolia logo to the list.

Thanks for such a great tool!


pixelastic:
Thanks. That was quick!

PR #144, opened 2015-10-09
Add Transloadit logo

kvz:
For convenience, here's how it looks among the others

On a side note, I noticed most companies post their full logo here (so I did as well) but it does result in a wide range of sizes:

```bash
$ du -hs * |gsort -h
4.0K	2gis.svg
4.0K	brigade.svg
4.0K	disqus.svg
4.0K	fingerproof.png
4.0K	millenium.svg
8.0K	cerner.png
8.0K	domain.svg
8.0K	paypal.svg
8.0K	schibsted.png
 12K	algolia.svg
 12K	alianza.png
 12K	atlassian.svg
 12K	computerbase.svg
 12K	the-native-web.png
 16K	qualitybath.png
 20K	geekpark.png
 24K	box.png
 24K	kahoot.png
 24K	transloadit.png
 24K	valor-software-sm.png
 28K	paypal.png
 32K	angular-formly.png
 44K	codinum.png
 44K	facebook.png
 48K	getable.png
 64K	simplifield.png
108K	windsorcircle.png
136K	cocktailjs.png
```

Some logos weighing in at 136KB as you can see.

I thought it would be bad to temper with the status quo it in this particular PR - but would you be interested in a PR that optimizes these images and sets a max dimension of e.g. 800px width? Ideal dimensions up for discussion ofc, I would imagine the companies still want crisp logos on retina for instance.

PPS
(we have started using ESLint in our open source project tus [here](https://github.com/tus/tus-js-client/pull/11) and will be using it for all our other js projects as well)

PR #145, opened 2015-10-10
Optimize logos

kvz:
As briefly discussed in #144, by merging this PR, the the total weight of company logos will go from `752K` to `312K`. The full change-set in byte sizes & dimensions can be reviewed in this commit: https://github.com/eslint/eslint.github.io/commit/b68920fd7b5663db53ead59533aa3ffa0ea0dff5

The weight can be brought down further by lowering `pngquant` quality in the `Makefile` and re-running `make optimize-logos` - but for our first swing I wanted to be on the conservative side and enjoy a 50% drop in sizes with no apparent visual impact.

The `Makefile` offers some crude dependency checking with suggestions that currently only target Homebrew for OSX. If there is a desire I could iterate on that and support Linux as well.

--

Unrelated, I'd like to elaborate on my brief proposal on Twitter to sponsor ESLint. Transloadit would like to match what your current banner campaign is bringing in, and sponsor an additional fee on top of that. We can also provide help with infra (expertise, servers, domains) should the need arise. In return we'd ask to remove banners and put up a plain [text link](https://dl.dropboxusercontent.com/s/6mgbvwz23g9pbg8/2015-10-10%20at%2009.32%202x.png) instead, for which I could send another PR. We have similar constructions with other open source projects that we rely on such as ImageMagick. It's good to do something back, and at the same time get the word out about our service. If it sounds interesting please shoot an [email](mailto:kevin@transloadit.com) - or ignore this and I won't bring it up again :blush:


ilyavolodin:
Hmm.. I'm not really sure how much good would this PR do for us, since only one or two people on ESLint team are using OSX. The rest of use are Windows users. So this would be unusable for most of the team.

kvz:
Ah, sorry if I was unclear.

 - This PR introduces optimized images. That will work on every OS.
 - As a bonus, I'm providing a script that can also optimize and report on newly added logos, should unoptimized ones be added in the future. It makes use of ImageMagick and pngquant. That could work on every OS.
 - As an added bonus, the Makefile provides install instructions should it detect that these prerequisite tools are not installed yet on the system. These suggestions are targeted at OSX.

So if the person in charge of optimizing future logos hits the warning "please brew install imagemagick" and they are a Linux user, they should interpret the message as "please apt-get install imagemagick", if they are a Windows user they should interpret it as "Please download and install ImageMagick" - but that's as far as OSX coupling goes. If we remove the suggestion, which is nothing more than a convenience, there is no OS(X) dependency.

ilyavolodin:
Sorry, lost track of this PR. Lets go with this, and if we get into trouble with ImageMagick later, we'll can adjust makefile later.

kvz:
:+1: When that happens please give a shout and maybe I can assist

kvz:
.. or just use the scripts again to optimize a new batch for logos and just send you another pr for that ;)

PR #157, opened 2015-11-13
Demo: Demo and Search Working (fixes #156)

### LINKED ISSUE #156: Demo: Demo does not work in Firefox
I get two errors in Firefox 42.0 (using Firebug console). Unfortunately, I'm not sure which of these are relevant. Sorry.

> ReferenceError: algoliasearch is not defined
>	var client = algoliasearch('XWXG7MEBSB', '653e00f423bee91f9863571eed16f2f5');
> 
> search.js (line 5, col 7)

And:

> TypeError: b is undefined
> ...}),k=d||new h({input:f,eventBus:j,hint:void 0===b.hint?!0:!!b.hint,minLength:b.m...

The second one seems to be from autocomplete.min.js, but not 100% sure.

The odd thing is, I actually get similar errors in Chrome, so... if my diagnostics are useless, I apologize.

xjamundx:
We had an extra jQuery in there which was causing conflicts this seems to work.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -31,7 +31,6 @@
   <script src="/js/vendor/jquery-1.10.2.min.js"></script>
   <script src="/js/vendor/bootstrap.min.js"></script>
   <script src="/js/vendor/anchor.min.js"></script>
-  <script src="https://cdn.jsdelivr.net/g/lodash@3.10.1,hogan.js@3.0.2,algoliasearch@3.8.1,autocomplete.js@0.12.0(autocomplete.min.js)"></script>
   <script src="/js/app/search.js"></script>
   <script>
     $(document).ready(function() {
--- a/js/app/demo.js
+++ b/js/app/demo.js
@@ -1,6 +1,6 @@
 // Code goes here
 
-require.config({
+requirejs.config({
     paths: {
         'text': '//cdnjs.cloudflare.com/ajax/libs/require-text/2.0.12/text',
         'orion': 'http://eclipse.org/orion/editor/releases/current/built-editor-amd.min'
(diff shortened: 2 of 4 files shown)


xjamundx:
Give me a second to test a bit more...

_includes/footer.html
```diff
@@ -31,8 +31,9 @@
   <script src="/js/vendor/jquery-1.10.2.min.js"></script>
   <script src="/js/vendor/bootstrap.min.js"></script>
   <script src="/js/vendor/anchor.min.js"></script>
-  <script src="https://cdn.jsdelivr.net/g/lodash@3.10.1,hogan.js@3.0.2,algoliasearch@3.8.1,autocomplete.js@0.12.0(autocomplete.min.js)"></script>
+  <script src="https://cdn.jsdelivr.net/g/lodash@3.10.1,hogan.js@3.0.2,algoliasearch@3.8.1,autocomplete.js@0.12.0(autocomplete.min.js),requirejs@2.1.20"></script>
   <script src="/js/app/search.js"></script>
+  <script src="/js/app/demo.js"></script>
```
ilyavolodin: This is included on every page of the site. Since we don't really need demo.js anywhere other than on the demo page, I don't think it should go here.

xjamundx: Yeah I'm still working it out....there are issues with how we now have a bunch of JS that loads here and we have conflicting JS loading earlier.

xjamundx:
Still more issues.....now with the parser page

xjamundx:
Moving the `algoliasearch` dep to the top worked. The alternative would be if jenkins supported a way for a partial (or whatever they're called) to set a block of JS that needs to be executed at the end.

js/app/demo.js
```diff
@@ -1,6 +1,6 @@
 // Code goes here
 
-require.config({
```
xjamundx: Appeared to be needed, though not sure why exactly. May be a conflict with a diff version of require being used somewhee or something, but this works

ilyavolodin:
Cool. Thanks @xjamundx

PR #174, opened 2016-01-12
Fix whitespace issue

mischah:
Replace example comment:

 ```js
/*eslint-env node, mocha */
 ```

with:

 ```js
/*eslint-env node, mocha */
 ```

**Reason:**
The example caused a linting error. 

```bash
1:1   error  Expected space or tab after "/*" in comment  spaced-comment
```

### DIFF
--- a/docs/user-guide/configuring.md
+++ b/docs/user-guide/configuring.md
@@ -129,7 +129,7 @@ Environments can be specified inside of a file, in configuration files or using 
 To specify environments using a comment inside of your JavaScript file, use the following format:
 
 ```js
-/*eslint-env node, mocha */
+/* eslint-env node, mocha */
 ```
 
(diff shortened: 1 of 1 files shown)


mischah:
I see that there are lots of other examples missing whitespace.

Example:
```js
/*global var1, var2*/
```

I could update the PR to fix them all in case you would accept the PR. Or ist that on purpose?

ilyavolodin:
Thanks for PR. Those comments only require space if the `spaced-comment` configured with `always`. I don't think there's a huge need to replace this style throughout the documentation.

mischah:
Okay. just wondered because `always` is the [default value](http://eslint.org/docs/rules/spaced-comment.html#options)

Fell free to close the PR :octocat:

PR #176, opened 2016-01-18
duplicate key

y8n:
`severity` is duplicate


nzakas:
Thanks, case you make this change in the ESLint repo? https://github.com/eslint/eslint/blob/master/docs/developer-guide/nodejs-api.md

y8n:
@nzakas  It seems like there is no such error in the ESLint repo.  
BTW. Thanks for ESLint  :+1:  and really expect for the official documentation in Chinese.

nzakas:
Ah ok, then I'll merge this so it goes on the live site. Thanks!

PR #178, opened 2016-01-20
Stop indexing v2.0.0 in search

pixelastic:
I've excluded all urls matching `/docs/2.0.0/` from the indexing, so
those docs should not be displayed in the searchbar.


pixelastic:
Sorry, bad merge. I'll reopen a new one.

PR #179, opened 2016-01-20
Stop indexing v2.0.0 in search

### LINKED ISSUE #177: Search bar defaults to returning results in 2.0.0
According to the latest [news](http://eslint.org/blog/2016/01/eslint-v2.0.0-beta.1-released/) the ESLint beta release isn't intended for production use. 

However:

- the documentation search bar always pulls up the 2.0 docs (even if I manually change the url to 1.0.0)
- there's nothing in the UI of the site to switch back to an older version, or even indicate which version is being viewed (it's visible in the browser url, but that's the only place)

This all seems counterintuitive, especially if production use of 2.0.0 is being discouraged.

pixelastic:
I've excluded all urls matching `/docs/2.0.0/` from the indexing, so those docs should not be displayed in the searchbar.

### Before
image

### After
image

Should fix #177

### DIFF
--- a/_plugins/search.rb
+++ b/_plugins/search.rb
@@ -49,6 +49,8 @@ class EslintCustomSearchHelper
     # Old docs
     return true if item[:url] =~ %r{^/docs/0.24.1/}
     return true if item[:url] =~ %r{^/docs/1.0.0/}
+    # Beta docs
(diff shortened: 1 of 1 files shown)


ilyavolodin:
Very nice! Thanks!

PR #180, opened 2016-01-21
Docs: Add icons for recommended and fixable rules

kaicataldo:
fixes eslint/eslint#4355

### DIFF
--- a/_layouts/doc.html
+++ b/_layouts/doc.html
@@ -3,7 +3,7 @@
 <main class="doc">
   <article class="container">
     {% include ad.html %}
-    {{ content }}
+    {{ content | replace: '(recommended)', '<span aria-label="recommended" class="glyphicon glyphicon-ok"></span>' | replace: '(fixable)', '<span aria-label="fixable" class="glyphicon glyphicon-wrench"></span>' }}
   </article>
 </main>
 {% include footer.html %}
--- a/styles/overrides.css
+++ b/styles/overrides.css
@@ -29,3 +29,9 @@ h6:target:before {
   height: 80px;
   width: 1px;
 }
+
+/* icon margins for rules page */
+.glyphicon {
+  margin: 0 2px;
+}
+


docs/rules/index.md
```diff
@@ -5,42 +5,45 @@ layout: doc
 <!-- Note: No pull requests accepted for this file. See README.md in the root directory for details. -->
```
ilyavolodin: Ah. Sorry, this file will have to be changed in ESLint main repository. It gets copied here automatically. I'm also not 100% sure that jekyll supports adding `span` into markdown.

kaicataldo: Is there concern re the jekyll support if I've tested it locally and ensured it's working? Pretty sure we can use some CSS hackery to put the icons in, but I felt like this approach would be easier for everyone involved.

ilyavolodin: We can definitely try it. Problem with local jekyll vs. github is that they are not exactly the same. But we can try and see if it works.

kaicataldo: Sounds like a plan, and I'm happy to try another approach if need be. Just so I'm clear, should I leave the style change in this PR and then open another PR to the ESLint main repository with the markdown changes?

ilyavolodin: Yes, css changes should stay in this PR and a new PR should be opened in main ESLint repository for md changes. Thanks!

nzakas: We also need descriptive text for accessibility purposes, an empty span isn't good enough. 

Also, I think we should keep the plain text versions in the ESLint repo and use the build script to replace "(recommended)" with the icon. That will make it easier to maintain.

_layouts/doc.html
```diff
@@ -3,7 +3,7 @@
 <main class="doc">
   <article class="container">
     {% include ad.html %}
-    {{ content }}
+    {{ content | replace: '(recommended)', '<span aria-label="recommended" class="glyphicon glyphicon-ok"></span>' | replace: '(fixable)', '<span aria-label="fixable" class="glyphicon glyphicon-wrench"></span>' }}
```
ilyavolodin: Very nice, I didn't realize you can do that. Did you verify this?

kaicataldo: Locally, yes! Here are [the docs](https://github.com/Shopify/liquid/wiki/Liquid-for-Designers#standard-filters) for Liquid filters if you're interested. Without the changes to index.md it replaces the instances of the words in the intro too, so if we want to implement the original design with the legend I'll have to make another small PR to the main repo.

ilyavolodin:
Ok, let's go with this for now, and we can add legend as a separate PR on the main repository.

kaicataldo:
Cool, I'll try to get to that later this evening

_layouts/doc.html
gyandeeps: One thing we missed is that we should add the `title`1 property to the span. It helps the enduser to know what it is when he hovers over the icon.

kaicataldo: Ah, you're right, that would be great. I can make another PR later for this later tonight.

gyandeeps: Awesome. Thanks a lot.

PR #182, opened 2016-01-22
Docs: Maintainer guide

nzakas:
@eslint/eslint-team  I've been meaning to write this up for a while, and finally got around to it. Please let me know any feedback you might have.


docs/maintainer-guide/issues.md
```diff
@@ -0,0 +1,58 @@
...
+## When an Issue is Opened
+
+When an issue is first opened, follow these steps:
+
+1. Is it clear what is being requested? If no, ask for more information. If yes, then:
+    * Label bug reports with the "bug" label (also use the "accepted" label if you can reproduce and verify the bug)
+    * Label requests for changes to existing features with the "enhancement" and "proposal" labels
+    * Label requests for completely new features with the "feature" and "proposal" labels
+    * Use an appropriate label for the part of the project the issue refers to
+1. If it's a bug, be sure the issue includes the following (ask for any missing information):
```
ilyavolodin: While I understand why you are asking for it, I don't think it's always necessary, if you can reproduce a bug with the information provided, it's redundant to ask for that information from the user. I think it can be very frustrating to new users to provide a lot of information, just to be asked to provide even more. I think this should be filed under "use your own judgement".

lo1tuma: Isn't this already covered by the eslint-bot anyway?

docs/maintainer-guide/issues.md
```diff
@@ -0,0 +1,58 @@
...
+1. If it's just a question (one that would normally go to the mailing list or chatroom), feel free to answer the question and close the issue when the conversation is over.
+1. If you can verify that the issue is a duplicate of another issue, mention this in a comment and close the issue. Be sure to reference the issue number of the issue they should refer to, such as, "Duplicate of #1234."
+1. Regardless of the above, always leave a comment. Don't just add labels, engage with the person who opened the issue by asking a question (request more information if necessary) or stating your opinion of the issue. If it's a verified bug, ask if the user would like to submit a pull request.
+
+**Note:** Don't add an "accepted" label to an issue unless it's a bug that you've been able to reproduce and verify (you're sure it's a bug). The "accepted" label will be added by a project admin if it's appropriate for the roadmap.
```
ilyavolodin: This is unclear. At some point, Governance model stated that reviewers are supposed to mark issues as accepted, has this been changed? (I'm fine with the change, just wanted to clarify).

docs/maintainer-guide/issues.md
```diff
@@ -0,0 +1,58 @@
...
+
+When a suggestion is too ambitious or would take too much time to complete, it's better not to accept the proposal. Stick to small, incremental changes and lay out a roadmap of where you'd like the project to go eventually. Don't let the project get bogged down in big features that will take a long time to complete.
+
+## When to Close an Issue
+
+There are several times when it's appropriate to close an issue:
+
+1. The issue is a duplicate of an existing issue
+1. The issue is just a question rather than a task to be completed
+1. The issue has been open for a year or more with no sign of progress
```
ilyavolodin: I think a year might be too long of a time frame. Half a year, maybe?

docs/maintainer-guide/pullrequests.md
```diff
@@ -0,0 +1,64 @@
...
+
+Pull requests are submitted frequently and represent our best opportunity to interact with the community. As such, it's important that pull requests are well-reviewed before being merged and that interactions on pull requests are positive.
+
+## Who Can Review Pull Requests?
+
+Anyone, both team members and the public, may leave comments on pull requests.
+
+## Who Can Approve Pull Requests?
+
+Only project reviewers or admins may used the "Merge" button to
```
ilyavolodin: extra "to"

docs/maintainer-guide/releases.md
```diff
@@ -0,0 +1,44 @@
...
+
+* Regular releases that follow [semantic versioning](http://semver.org/) and are considered production-ready.
+* Prereleases that are not considered production-ready and are intended to give the community a preview of upcoming changes.
+
+## Setting up a Release Environment
+
+In order to run a release, be sure that:
+
+1. You have [eslint.github.io](https://github.com/eslint/eslint.github.io) checked out into the same directory as the project (such as ESLint itself). Some projects automatically update the website.
+1. You are logged in to npm. If not, [log in](https://docs.npmjs.com/cli/adduser).
```
ilyavolodin: Make sure that git line ending set to "Unix":-)

docs/maintainer-guide/issues.md
```diff
@@ -0,0 +1,58 @@
...
+
+## When an Issue is Opened
+
+When an issue is first opened, follow these steps:
+
+1. Is it clear what is being requested? If no, ask for more information. If yes, then:
+    * Label bug reports with the "bug" label (also use the "accepted" label if you can reproduce and verify the bug)
+    * Label requests for changes to existing features with the "enhancement" and "proposal" labels
+    * Label requests for completely new features with the "feature" and "proposal" labels
+    * Use an appropriate label for the part of the project the issue refers to
```
btmills: It might be helpful to list at least some of "build", "cli", "core", "documentation", "infrastructure", "rule", and "website" as examples here.

docs/maintainer-guide/releases.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Regular Releases
+
+You can run a regular release with the following command:
+
+```
+$ npm run release
+```
+
+The [release tool](https://github.com/eslint/eslint-release) will inspect the commit history and determine the correct next version. It will also handle publishing to both GitHub and npm with the correct names and tags. Each project may have additional steps for pushing a release, but the basic steps are the same for all.
```
ilyavolodin: I know you mentioned that each project may have additional steps, but updating release notes on github will probably be common (until functionality is added to eslint-release)

docs/maintainer-guide/issues.md
lo1tuma: If the issue is a bug or describing technical debt I would not apply this rule. Because we would loose track of those problems then.

docs/maintainer-guide/issues.md
ilyavolodin: It is. People ignore bots:-) I think that implies that if they didn't answer bot's question, you should ask again.

docs/maintainer-guide/issues.md
```diff
@@ -0,0 +1,58 @@
...
+    * The actual ESLint output (not a summary, have them copy-paste it)
+    * What they expected to happen
+1. If it's an enhancement or feature request, be sure the issue mentions the primary use case (the problem they are trying to solve) and not just a suggestion for a technical change (which needs to be a solution to a problem). If the use case isn't clear, ask for more information. Frequently, "What problem are you trying to solve?" can clarify that.
+1. If it's just a question (one that would normally go to the mailing list or chatroom), feel free to answer the question and close the issue when the conversation is over.
+1. If you can verify that the issue is a duplicate of another issue, mention this in a comment and close the issue. Be sure to reference the issue number of the issue they should refer to, such as, "Duplicate of #1234."
+1. Regardless of the above, always leave a comment. Don't just add labels, engage with the person who opened the issue by asking a question (request more information if necessary) or stating your opinion of the issue. If it's a verified bug, ask if the user would like to submit a pull request.
```
ilyavolodin: This one is tough. Are you saying that if you are not knowledgeable enough to answer the question, or respond to a request you shouldn't label it? I know that I do triage run in the morning, and even if I find a new issue with the question I can't answer, I try to label it, for future references. Should I not do that, and leave labeling to somebody who can?

docs/maintainer-guide/issues.md
btmills: This tripped me up too. The [governance model](http://eslint.org/docs/developer-guide/governance) has an Administrator role, but it sounds like "project admin" here (and on the line 40 below) refers to either Project Lead or Reviewer?

ilyavolodin:
In general looks good to me. However, there is one general issue that isn't addressed. You keep talking about "scope of the project", however, problem is, that each one of us has a different notion as to what exactly that means. The only one who has the "right" notion is you. We don't have a very detailed roadmap, and while I think everyone on the team has a good understanding where we want to get, I'm sure that it varies in details among team members.

docs/maintainer-guide/issues.md
nzakas: @ilyavolodin I tend to use a year because it's hard to argue with. "Really, it's been a year and no movement?"

@lo1tuma if something is that important then it should have somehow bubbled up within a year. Anything that hangs around this long can't possibly be important enough to keep open.

docs/maintainer-guide/releases.md
nzakas: Doesn't even matter now thanks to the build script. :)

docs/maintainer-guide/issues.md
nzakas: This applies to all projects, not just ESLint, so I'm not sure how specific we should get here. Open to suggestions.

docs/maintainer-guide/releases.md
nzakas: Yeah, that should get built into the release tool.

docs/maintainer-guide/issues.md
nzakas: The overall point is that if you take the time label it, leave a comment as well. Just labeling it doesn't tell anyone what's going on...did you just not have time to comment? Do you not have anything to say? I have a lot of trouble figuring out what to do when I see someone has labeled a issue but hasn't left a comment. Even if you just say, "looks like a bug, we'll take a look", that's better than nothing.

nzakas:
I'm not even sure I have the "right" notion of this, as it's fluid and varies from project to project (note that this maintainer guide is *not* specific to ESLint, but rather applies to all projects in the ESLint org). Such info, if anywhere, should probably be in the readme rather than here.

As for what the scope of ESLint is - we've always kind of come up with that as we've gone along. For instance, at one point we all agreed that autofixing was out of scope, but then it kept getting requested so it became in scope. Same with JSX. So, I'm not sure what you'd like along these lines, but if you can give me some more specific examples, maybe we can come up with something together.

docs/maintainer-guide/issues.md
ilyavolodin: Don't get me wrong, I'm not saying you should change this paragraph, more like asking for directions. https://github.com/eslint/eslint/issues/5033 is a good example. If I would see this issue in the morning, I would label it with `question` and not leave any comments, because I'm not familiar enough with ESTree spec. I wouldn't want to ask a question either, since I'm pretty sure you and @mysticatea would know exactly how to answer it, without needing clarification. So should I just not label an issue like that, and leave it for somebody else?

ilyavolodin:
I think having a more flashed out roadmap would probably help. It will not answer all of the questions, but we defiantly help set the direction. And having information about a few milestones beyond the next one would help as well. Before we release 1.0 we had milestones for like 8-9 releases in advance. I understand that it would be significantly harder to do now, but maybe 2-3 is possible?

docs/maintainer-guide/pullrequests.md
IanVS: Nit: "used" should be "use"

docs/maintainer-guide/issues.md
btmills: Given that, I think this suggestion doesn't apply - anything I can think of would be far too verbose and distracting.

docs/maintainer-guide/pullrequests.md
```diff
@@ -0,0 +1,64 @@
...
+
+In any of these cases, please be sure to leave a comment stating why the pull request is being closed.
+
+### Example Closing Comments
+
+If a pull request hasn't been updated in 30 days:
+
+> Closing because there hasn't been activity for 30 days. If you're still interested in submitting this code, please feel free to resubmit.
+
+If a pull request submitters isn't willing to follow project guidelines.
```
IanVS: Nit: "submitters" should be "submitter".

nzakas:
@ilyavolodin a roadmap would be great. The reason I only do them one at a time is because:

1. Maintaining a roadmap is hard as things change. I did that to start the project and it ended up an unmaintained list. 
1. I never know what sort of contributions to expect and I don't want to promise things that just won't happen.
1. A lot of the time, I have no idea what's coming up. Most of the time, the roadmaps come together from the issues that are filed and the suggestions the team makes.

I've tried to at least get the next version roadmap by opening those "wish list" issues periodically. If you have suggestions for how to improve that or what else can be done, definitely let me know.

docs/maintainer-guide/issues.md
lo1tuma: @nzakas Good point. Apart from that the issue could also be re-opened if it becomes important someday.

mysticatea:
Looks good to me.

docs/maintainer-guide/index.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Maintainer Guide
+layout: doc
+---
+
+# Maintainer Guide
+
+This guide is intended for those who are part of any ESLint project team.
```
alberto: Shouldn't governance doc be moved here too?

docs/maintainer-guide/issues.md
alberto: Yeah, I think using the same terminology would clear things up.

docs/maintainer-guide/issues.md
```diff
@@ -0,0 +1,58 @@
...
+1. **Be inquisitive.** Ask questions on the issue whenever something isn't clear. Don't assume you understand what's being reported if there are details missing. Whenever you are unsure, it's best to ask for more information.
+1. **Not all requests are equal.** It's unlikely we'll be able to accommodate every request, so don't be afraid to say that something doesn't fit into the scope of the project or isn't practical. It's better to give such feedback if that's the case.
+1. **Close when appropriate.** Don't be afraid to close issues that you don't think will be done, or when it's become clear from the conversation that there's no further work to do. Issues can always be reopened if they are closed incorrectly, so feel free to close issues when appropriate. Just be sure to leave a comment explaining why the issue is being closed (if not closed by a commit).
+
+## When an Issue is Opened
```
alberto: We use the term `triage` for this. It would be nice to explain it here (or at least refer to it), since it is not a very common term.

docs/maintainer-guide/issues.md
```diff
@@ -0,0 +1,58 @@
...
+
+If you're a project admin, then you'll need to evaluate incoming requests for inclusion in the formal roadmap. You should label an issue as "accepted" when all of the following are true:
+
+1. The feature or enhancement is in scope for the project and should be added to the roadmap
+1. You are committed to including the change within the next year
+1. You are reasonably certain of who will do the work
+
+When a suggestion is too ambitious or would take too much time to complete, it's better not to accept the proposal. Stick to small, incremental changes and lay out a roadmap of where you'd like the project to go eventually. Don't let the project get bogged down in big features that will take a long time to complete.
+
+## When to Close an Issue
```
alberto: Mention who can do that.

docs/maintainer-guide/releases.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Prereleases
+
+You can run a prerelease with the following command:
+
+```
+$ npm run prerelease -- <release version>
+```
+
+You need to replace `<release version>` with a prerelease version. Prelease versions are in the form 0.0.0-tag.0, such as 2.0.0-beta.1. For example:
```
alberto: Replace the term `tag` with `identifier`, which is what the semver spec uses.

docs/maintainer-guide/pullrequests.md
nzakas: Grammar is hard :(

docs/maintainer-guide/issues.md
nzakas: I see what you mean. My concern is that the person submitting the issue doesn't get any notification when a label is added, but does when a comment is left (so they know someone has looked at it). Maybe we need another label for "needs review"? So you can categorize but still mark as needing a response?

docs/maintainer-guide/index.md
nzakas: Good suggestion, absolutely.

docs/maintainer-guide/issues.md
nzakas: I'll add it.

docs/maintainer-guide/releases.md
nzakas: Nice catch

docs/maintainer-guide/issues.md
BYK: May be make the bot do this automatically with periodic sweeps over open issues?

nzakas:
Just an update on this: I'm going to rewrite this to be specific to ESLint. That's where we really need the docs, so I was being a bit silly by trying to make it work for every project. :)

nzakas:
Updated with more specifics and ESLint project focus.

docs/maintainer-guide/issues.md
```diff
@@ -0,0 +1,82 @@
...
+
+When a issue is opened, the bot will automatically apply the "triage" label. Issues labeled with "triage" are the ones that need to be looked at by team members to determine what to do next.
+
+The steps for triaging an issue are:
+
+1. Is it clear what is being requested?
+  * No: add the "needs info" label to the issue. The bot will add a comment asking for more information. You don't need to comment any further until the person who opened the issue responds with the information requested from the bot.
+  * Yes:
+    * Label questions with the "question" label
+    * Label bug reports with the "bug" label (also use the "accepted" label if you can reproduce and verify the bug)
```
ilyavolodin: Should we not label unverified bugs with `proposal` label as well?

docs/maintainer-guide/issues.md
```diff
@@ -0,0 +1,82 @@
...
+  * **New Rules:** if you are willing to champion the rule (meaning you believe it should be included in ESLint core), then add a comment saying you will champion the rule and add an "accepted" label. As the champion, it's your job to guide the rule creation process until it's complete, so only champion a rule that you have time to implement or help another contributor implement it.
+  * **Duplicates:** if you can verify the issue is a duplicate, add a comment mentioning the duplicate issue (such as, "Duplicate of #1234") and close the issue.
+1. Regardless of the above, always leave a comment. Don't just add labels, engage with the person who opened the issue by asking a question (request more information if necessary) or stating your opinion of the issue. If it's a verified bug, ask if the user would like to submit a pull request.
+
+**Note:** Don't add an "accepted" label to an issue unless it's a bug that you've been able to reproduce and verify (you're sure it's a bug). The "accepted" label will be added by a project admin if it's appropriate for the roadmap.
```
ilyavolodin: This goes against a paragraph above about championing new rule.

docs/maintainer-guide/issues.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Evaluating an Issue
+
+If you're a project admin, then you'll need to evaluate incoming requests for inclusion in the formal roadmap. You should label an issue as "accepted" when all of the following are true:
+
+1. The feature or enhancement is in scope for the project and should be added to the roadmap
+1. You are committed to including the change within the next year
+1. You are reasonably certain of who will do the work
```
ilyavolodin: What about `accepted`+`help needed`? As in you know that somebody from outside of the team will need to do the change/feature. We currently do this all the time.

docs/maintainer-guide/pullrequests.md
```diff
@@ -0,0 +1,64 @@
...
+## Who Can Approve Pull Requests?
+
+Only project reviewers or admins may use the "Merge" button to merge in changes.
+
+## Reviewing a Pull Request
+
+When a pull request is opened, follow these steps:
+
+1. Has the submitter signed a CLA? If not, please ask them to do so.
+1. Is the commit message summary in the correct format? Double-check that the tag ("Fix:", "New:", etc.) is correct based on the issue. Documentation-only pull requests do not require an issue.
```
ilyavolodin: Should probably mention that bot will check for those and leave a message, if it finds problems.

nzakas: Good call.

docs/maintainer-guide/issues.md
nzakas: Yup, I'll add that.

docs/maintainer-guide/issues.md
nzakas: I don't know. I usually don't, because "proposal" seems like the wrong word when it comes to bugs. I added "proposal" for features and enhancements, but we can use it on bugs if you think it makes sense.

docs/maintainer-guide/issues.md
nzakas: > At some point Governance model stated that Reviewers are supposed to put accepted label.

I don't believe this is true. You can check the [full history](https://github.com/eslint/eslint.github.io/commits/master/docs/developer-guide/governance.md). I initially asked that only I add "accepted", and the document reflects that. Later, we went to anyone can add "accepted" for verified bugs, and then after that, anyone can add "accepted" for rules they will champion.

ilyavolodin: Sounds good to me. Just wanted to clarify this point, I'm OK with leaving marking `accepted` features to you.

docs/maintainer-guide/releases.md
```diff
@@ -0,0 +1,39 @@
+---
+title: Managing Releases
+layout: doc
+---
+
+# Managing Releases
```
alberto: Is it worth adding who can do this?

docs/maintainer-guide/pullrequests.md
```diff
@@ -0,0 +1,72 @@
...
+Ideally, the first reviewer to comment on the pull request should be the one to merge it. However, if you don't feel comfortable doing so for any reason, just leave a comment saying "LGTM but would like someone else to verify".
+
+Be sure to say thank you to the submitter before merging, especially if they put a lot of work into the pull request.
+
+## When to Close a Pull Request
+
+There are several times when it's appropriate to close a pull request without merging:
+
+1. The pull request addresses an issue that is already fixed
+1. The pull request hasn't been updated in 30 days
```
alberto: Maybe out of the scope of this issue, but these two tasks seem ideal for eslintbot too.

nzakas: Yeah, a bit tougher to automate, but I'll look into it.

docs/maintainer-guide/releases.md
nzakas: Yeah, good point

PR #188, opened 2016-01-29
Fixed typo, added rules mentioned in previous post

mjomble:
There was a backtick missing on the one-var-declaration-per-line line.
The other missing rules were the same ones that had to be fixed in the previous post by #184


mysticatea:
Thank you very much!

ilyavolodin:
Thanks. The other rules that you mentioned were added prior to beta-2 and are noted in the beta-1 release notes and before.

mysticatea:
@ilyavolodin The post contains contents (highlights and commits) of beta-2 and before. But it's lacking several new rules in the new rule list.

ilyavolodin:
Sounds good in that case.

nzakas:
Thanks, I end up doing this by hand, so it's always a bit dicey.

PR #199, opened 2016-02-14
Update parser demo for espree 3.0 (fixes #196)

### LINKED ISSUE #196: Update parser demo
The parser demo needs to be updated with the correct options for the new version of Espree. So, it should have ecmaVersion as an option and allow you to specify `sourceType` as "script" or "module".

@xjamundx I couldn't quite tell how the parser demo was built, can you take a look (and also, can you add some docs in the readme about how to build the parser demo files)?

xjamundx:
screen shot 2016-02-15 at 9 45 23 am

Fixes everything. Also displays the current Espree Version.  The only issue I have now is why does the `eslint` in this repo include the outdated `espree` version `2.2.5`? Other than that everything seems to work fine!

### DIFF
--- a/js/app/parser/index.built.js
+++ b/js/app/parser/index.built.js
@@ -7,32 +7,67 @@ var lib = require('./lib');
 var editor = document.querySelector('#editor');
 var results = document.querySelector('#results');
 var controls = document.querySelector('#configuration');
-var optionsControl = document.querySelector('.options .list');
 
-function handleBodyClick(event) {
-	if (event.target.classList.contains('option-checkbox')) {
-		toggleControl(event.target.id);
+function handleSelectChange(property, el) {
+	var selected = el.value;
+	config[property] = selected;
+}
+
+function toggleOption(option, property) {
+	if (property) {
+		config[property][option] = !config[property][option];
+	} else {
+		config[option] = !config[option];
 	}
 }
 
-function toggleControl(id) {
-	config[id] = !config[id];
-}
-
-function setupControls() {
-	// setup basic handler for all clicks
-	document.body.onclick = handleBodyClick;
+function drawCheckboxes(config, property) {
+	var el = document.querySelector('.' + (property || 'options'));
 
 	// convert the list of options into checkboxes
 	var checkboxes = Object.keys(config).filter(function(option) {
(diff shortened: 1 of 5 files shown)


nzakas:
Looks nice!

platinumazure:
@xjamundx @nzakas I'm super excited to see this land-- what's left, and can I help?

gyandeeps:
I think we should also mention what version of `eslint` is the demo running. Its very helpful in cases like what we have right now. Currently it doesnt say that its using eslint 1.x

xjamundx:
@gyandeeps this is just the espree demo. not sure the eslint version is as relevant.

AFAIK this is done, but it needs some type of build process to get the latest espree version.

gyandeeps:
oops sorry about that.
Should we add the version on the parser demo also?

xjamundx:
but yeah i added the espree version!

gyandeeps:
:+1:

ilyavolodin:
@xjamundx Should we modify ESLint release script to also push latest Espree? Or maybe modify Espree release script directly?

xjamundx:
@ilyavolodin I'd like to get to the point where this thing had its own release script, which would open the door to us updating our tooling a bit and (getting rid of require.js for example).

As far as this PR, I think it's ready to go now, it should be backwards compatible with the current version AFAIK, and then whenever the website is updated with the new eslint/espree version it should continue working.

ilyavolodin:
@xjamundx I merged this in, and it works, but it looks like there are bugs in implementation. Simple code like `let foo = bar;` only parses when `module` is selected. With `script` it always returns `Unexpected token foo`. Also when `module` is selected, it parses even if I select `ecmaVersion` 3 (I have no idea what's supposed to happen with that setup, but I don't think it should parse, unless module forces es6 no matter what was selected). Also, it looks like it tries to parse on the timeout. Even if I don't type anything in the textbox, it periodically re-parses my code (which I noticed cause you left console.logs in or maybe it's coming from Espree directly?)

xjamundx:
Thanks for the note. Is it still on the older espreee version? I guess the
options are for espree 3 so that would explain the weird behavior?

PR #201, opened 2016-02-14
Lighten code background to improve readabilty

francoisromain:
The light blue background on the code is too dark and not easy to read.


ilyavolodin:
@francoisromain Thanks for PR. Unfortunately less files are not used for anything, as we have no compilation step (it was an oversight in the begging for the site, where somebody compiled all of the less files locally and just checked in resulting css file). So this change will not have any affect on the css of the site.

francoisromain:
Should I do the change inside `main.936b04d2.css`?

ilyavolodin:
Yes, unfortunately that's the only way to do it now, since there have been already manual changes to that file, we can't regenerate it by compiling less anymore.

PR #207, opened 2016-02-24
Remove space at the beginning of imported path.

eventualbuddha:


nzakas:
Sorry, this doc actually comes from the main ESLint repo. https://github.com/eslint/eslint/blob/master/docs/rules/no-restricted-imports.md

Changes need to be made there rather than here.

PR #210, opened 2016-03-04
Proposal: Increase scope of committership

nzakas:
Recently, the ESLint Reviewers concluded that it would be good to give ESLint Committers more authority in the project. This pull request proposes:

1. Committers should be allowed to merge some pull requests (bug fixes, docs, build-related, upgrades) on their own so long as they are non-breaking.
1. Committers should be allowed to label an issue as "accepted" if it is a bug fix, a new rule, or a non-breaking change to an existing rule. I think this change is good because it didn't make sense to allow Committers to accept new rules but not changes to existing rules.

Please let me know your thoughts. @eslint/eslint-team

### DIFF
--- a/docs/maintainer-guide/governance.md
+++ b/docs/maintainer-guide/governance.md
@@ -31,11 +31,10 @@ Committers:
 
 * Are expected to work on public branches of the source repository and submit pull requests from that branch to the master branch.
 * Are expected to delete their public branches when they are no longer necessary.
-* May submit small changes (documentation updates, changes to tests or code comments, configuration changes) without pull requests.
-* Must submit pull requests for any non-trivial changes.
+* Must submit pull requests for all changes.
 * Have their work reviewed by Reviewers before acceptance into the repository.
-* May label issues as they are submitted ("accepted" label should only be added for bugs, and only if the committer verified the bug as valid).
-* May close issues if they are duplicates of already resolved issues.
+* May label and close issues (see [Managing Issues](issues.html))
+* May merge some pull requests (see [Managing Pull Requests](pullrequests.html))
 
 To become a Committer:
 
--- a/docs/maintainer-guide/issues.md
+++ b/docs/maintainer-guide/issues.md
@@ -52,13 +52,15 @@ The steps for triaging an issue are:
 1. Next steps:
   * **Questions:** answer the question and close the issue when the conversation is over.
   * **Bugs:** if you can verify the bug, add the "accepted" label and ask if they would like to submit a pull request.
-  * **New Rules:** if you are willing to champion the rule (meaning you believe it should be included in ESLint core), then add a comment saying you will champion the rule and add an "accepted" label. As the champion, it's your job to guide the rule creation process until it's complete, so only champion a rule that you have time to implement or help another contributor implement it.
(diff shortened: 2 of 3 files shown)


gyandeeps:
:+1:

ilyavolodin:
While I have absolutely no problem with this change, my question is what's the difference between committer and reviewer at this point? Is it that reviewers can merge any pull request, and committers can only merge non-breaking?

nzakas:
> Is it that reviewers can merge any pull request, and committers can only merge non-breaking?

Not quite, but the primary difference is that committers can merge some pull requests now. It's actually a subset of non-breaking changes, not all.

BYK:
Kind of agree with @ilyavolodin. :+1: for accepting things, :-1: for merging.

nzakas:
@BYK why :-1: for merging?

mysticatea:
:+1:

alberto:
Is the intention to limit merges to trivial PRs?

BYK:
@nzakas that blurs the line between committers and reviewers a bit too much if you ask me. As a committer, if I can do merges too what separates me from the people who has spent much more time on the project? How about "they can merge non-breaking changes at least one reviewer says 'LGTM'`?

nzakas:
@alberto yes.

@byk I explicitly named what type of PRs committers can merge both in the original description of this PR and in the changes files. There is just a small subset that committers can merge on their own. Please read those over.

The overall goal here is to remove bottlenecks by increasing the number of things committers can do.

BYK:
@nzakas - I already did :) I think it would confuse me since now I have to keep in mind which patches I can merge and probably err on the safe side and not merge anything at all.

alberto:
@BYK The way I understand it, it would be only for trivial PRs consisting of a few lines of code, in order to get PRs merged faster and alleviate some of the work of the reviewers to allow them to focus on the non trivial ones. I'm :+1: for that

nzakas:
@byk you wouldn't have to merge anything, it's optional. If you don't want to, you don't have to.

BYK:
Alright, if I'm the only one concerned I'll yield :)

nzakas:
@btmills @IanVS @lo1tuma @xjamundx thoughts?

IanVS:
I'm a little nervous with so many people being able to accept new rules.  It seems to me that this should be a responsibility of the project leader(s) (after getting input from the team if desired).

nzakas:
@IanVS that hasn't changed. Committers have always been able to accept new rules, the only change here is that you'll now also be able to accept modifications to the rules.

IanVS:
Ah okay, I didn't look closely enough. Sounds good to me.

xjamundx:
Great I didn't realize we could even label issues. 0 concerns from me.

btmills:
No objections here!

nzakas:
By the way, just in case it's not clear, this is about adding *capabilities* for committers. You are not required to do anything you're not comfortable with, so of you are unsure of accepting something, or you want someone else to review a PR because you don't feel comfortable doing it, that's OK! Just leave a comment asking for someone else to help (which I think we all basically do anyway).

nzakas:
OK, I think we've reached consensus here, so merging.

PR #215, opened 2016-03-06
Upgrading Jekyll to 3.0.0 (fixes #194)

### LINKED ISSUE #194: Update site for Jekyll 3
I already got an email notification saying that redcarpet will be disabled on May 1st. We should update the site to use kramdown and see if we can also enable any of the new features.

ilyavolodin:
This should PR switches us from redcarpet to kramdown. I also added code highlighting while I was at it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,9 +1,9 @@
 source 'https://rubygems.org'
 
-gem 'jekyll', '~> 2.5.3'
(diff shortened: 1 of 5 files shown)


ilyavolodin:
Merging. Will revert if stuff breaks.

PR #217, opened 2016-03-07
Docs: Point to gitter chat as well as mailing list in change request docs

platinumazure:
Feel free to close if you would prefer people go to the mailing list, but I think one or two other places put the mailing list and chat on the same level, as "media we would prefer that people use to ask questions".


nzakas:
Fine by me.

PR #220, opened 2016-03-08
Update Algolia config

### LINKED ISSUE #216: Algolia push fails in Travis
I just upgraded site to use kramdown and jekyll 3.0, but now algolia push fails with every Travis build with the following message:
```
Bundled gems are installed into ./vendor/bundle.
22.27s$ bundle exec jekyll algolia push
Configuration file: /home/travis/build/eslint/eslint.github.io/_config.yml
     Build Warning: Layout 'none' requested in feed.xml does not exist.
jekyll 3.0.3 | Error:  undefined method `to_sym' for nil:NilClass
The command "bundle exec jekyll algolia push" exited with 1.
```
@pixelastic any chance you can help with this? I really don't know anything about ruby.

Also two other search related questions that I had were:
* Is there a way to tie in Algolia search with google analytics somehow?
* Search results seem to be adding `.html` extension to all search results, which confuses google with a bunch of duplicate pages, any way to remove that?

pixelastic:
Fixes #216

Seems likes the move to kramdown caused the root issue. Kramdown seems
to have trouble converting markdown that looks like this:

```
<!-- HTML comment -->
# Title
```

Instead of creating the `h1`, it kept the `#`. Adding a newline
between the comment and the title fixes the issue, but I actually
fixed it in another way.

I realize that the current setting was indexing every rule, for every
version. I felt this was not the best way to do it, as this requires
blacklisting versions everytime you would add a new one. Instead,
I chose to only index the rules in `/docs/rules`, without any version
specified.

Turns out that the files in this directory do not have the HTML
comment bug..

I also updated the underlying gem and fixed a edge-case with the Rule
index page.

In the end it decreased the number of records from ~9300 to ~5200, by
removing all duplicate rules of various versions.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
 gem 'jekyll', '~> 3.0.0'
 
 group :jekyll_plugins do
-  gem 'algoliasearch-jekyll', '~> 0.5.1'
+  gem 'algoliasearch-jekyll', '~> 0.7.0'
   gem 'jekyll-sitemap', '~> 0.10.0'
   gem 'jekyll-redirect-from', '~> 0.9.1'
 end
--- a/_plugins/search.rb
+++ b/_plugins/search.rb
@@ -46,11 +46,8 @@ class EslintCustomSearchHelper
     return true if item[:url] == '/docs/user-guide/integrations/index.html'
     return true if item[:url] == '/index.html'
     return true if item[:url] == '/parser/index.html'
-    # Old docs
-    return true if item[:url] =~ %r{^/docs/0.24.1/}
-    return true if item[:url] =~ %r{^/docs/1.0.0/}
-    # Beta docs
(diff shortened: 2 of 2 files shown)


ilyavolodin:
Nice! Thanks so much for looking at this! I fixed the rules under `/docs/rules` directory to have an extra space after comments due to the issue you described above, but since we do not republish previous version of rules, I couldn't fix it there. All new docs should also have a whiteline after comments now.

ilyavolodin:
Hmmm.. Ok, I'm confused. It looks like I merged this PR, but now it doesn't show up in the list of merged pulls and I also get a notification to submit new PR. Oh! I see, you submitted it against jekyll branch that I created. Ok, I'll remerge it back onto master.

pixelastic:
Yep, I wasn't sure how to reproduce the bug so I tried on the jekyll branch. Sorry for the confusion :)

PR #222, opened 2016-03-09
Update no-confusing-arrow.md

2color:
Correct the example of how it could be misunderstood.

As shown in https://groups.google.com/forum/#!topic/eslint/pHi9AUQ8hFk

### DIFF
--- a/docs/2.0.0/rules/no-confusing-arrow.md
+++ b/docs/2.0.0/rules/no-confusing-arrow.md
@@ -13,7 +13,7 @@ Here's an example where the usage of `=>` could be confusing:
 // The intent is not clear
 var x = a => 1 ? 2 : 3
 // Did the author mean this
-var x = function (a) { return a >= 1 ? 2 : 3 }
+var x = function x(a) { return 1 ? 2 : 3; };
 // Or this
 var x = a <= 1 ? 2 : 3
 ```


pedrottimark:
Thank you for following through to correct the mistake that you discovered.

This file in eslint.github.io is copied during the build process from a source file in **eslint**:

https://github.com/eslint/eslint/blob/master/docs/rules/no-confusing-arrow.md

Therefore, you need to submit a PR in the **eslint** project instead 🙃

Here are two thoughts to consider about the change:
* `var x = function (a) { return 1 ? 2 : 3; };` seems like a minimal change. Adding a function name **x** as `function x(a)` could distract from the main point of a function expression corresponding to the arrow function.
* As you noticed, there is inconsistency in semicolon style of the example code. @ilyavolodin Can you advise us what is the preferred style, so @2color can make it consistent here and so I can add that detail to the guidelines?

1. Make sure to sign the Contributor License Agreement (CLA) at http://eslint.org/cla/
2. As described in http://eslint.org/docs/developer-guide/contributing/pull-requests#step2 begin your commit message with Docs:

2color:
Thanks for the informative answer @pedrottimark. I'll take care of that in the next day or two.
And would be great if @ilyavolodin could advise on the preferred style.

ilyavolodin:
@2color Thanks for the issue and PR. ESLint uses semicolons everywhere in our codebase. So I think they should be added to the examples. I think we can assume that anyone who uses ASI will understand examples with semicolons, but we can't assume the opposite.

PR #226, opened 2016-03-15
Adding highlights for code examples

### LINKED ISSUE #108: Hard to separate working blocks from failing ones
I know the docs are autogenerated. But maybe there's a way to do that:

Currently, every block is formatted the same way. It would be much clearer to use some sort of color coding to separate the "good" examples from the ones that fail tests. A good example for color coding is found on [bootstrap](http://getbootstrap.com/components/#alerts-links).

Would be looking like that:
rules

ilyavolodin:
Partially fixes #108 only for the rules that were changed to the new style. Also it doesn't fix all of the examples even on those: http://eslint.org/docs/rules/no-constant-condition for example, has multiple separate example blocks for both correct and incorrect examples.


pedrottimark:
Nice. Thank you for investing time to test and catch the cases that do not work!

Follow-up by Mark:
* Work on more careful CSS selectors for multiple blocks of example code.
* Add to #108 pictures of alternatives to red/green color contrast for better accessibility.

nzakas:
It might be easier to include something like `<div class="good">` directly in the MD files. We'd have to look out for it in code reviews, but seems like a minimal effort to ask people to do.

pedrottimark:
@ilyavolodin So we can view changes to **main.css** on GitHub, what do you think if each rule is on its own line? Also the opening `@media … {` and closing `}` of media query.

File size increases only 1.5% but I already see at least one redundant (part of) a rule:

```
code,kbd,pre,samp{font-family:monospace,serif;}
code,kbd,pre,samp{font-family:Menlo,Monaco,Consolas,"Courier New",monospace}
```

If you think this is worthwhile, I can improve a very short JavaScript program running in Node.js to make the one-time change.

ilyavolodin:
@pedrottimark Thing is, main.css is generated by compiling bootstrap. Unfortunately, the person that did that, never checked in original modified bootstrap files. In reality, we just need somebody to pull Bootstrap 3 and skin it to look the same as our side, and then check it into the repository (and also add an npm script to compile it back into main.css file). That would be much more readable. I was hoping to do that when we do site redesign.

pedrottimark:
@nzakas Good to know that is an option. A selector is more subtle than I supposed 😳

Were you thinking the div encloses an example sentence paragraph and all its example code blocks?

pedrottimark:
@ilyavolodin That makes sense. Until then, I will use a replacement script to make a local version for faster study and editing so I don’t overlook anything important.

nzakas:
I was thinking just put the `<div>` around the code block, so:

```html
<div class="good">
`` ``js
// js code
`` ``
</div>
```

ilyavolodin:
@nzakas Are you opposed to merging it in, and seeing how it works for now (reverting it is always on the table)?

ilyavolodin:
I actually really like example with icons (no color).I think it conveys idea very well, and easy to read.

pedrottimark:
Agree on no color. All text aligned at same margin is much more readable for my eyes, but agreed with @nzakas that icon in margin once in a while was a problem, so the brainstorm was to make it a pattern.

ilyavolodin:
Ok. I'll make modifications for icons. Let's try it out in prod and see how it works. If something is off, we can always roll it back.

ilyavolodin:
OK, updated with icons for both code examples as well as "Fixable". Going to merge this in and see how it looks.

PR #227, opened 2016-03-18
Docs: Change to rule acceptance

nzakas:
This is a proposal for changing our procedure for accepting rules and rule changes. The general feeling amongst reviewers is that we have too many rules, and we're also spending a lot of time evaluating and adding new rules and rule changes.

This proposes that:

1. We only accept new rules if they are absolutely necessary for ESLint to be complete. Otherwise, we should suggest people create custom rules.
1. If a rule is so important that we accept it, then the rule must be implemented by the person who suggested it or a designee of that person. It will not be done by an ESLint team member.
1. Rule changes follow the same process: it must be important, and if we accept it, then the person who suggested it or a designee must implement it. It will not be done by an ESLint team member.

That basically means any new rule or rule change that is accepted must be labeled with "help wanted" immediately. 

Please review @eslint/eslint-team

### DIFF
--- a/docs/developer-guide/contributing/rule-changes.md
+++ b/docs/developer-guide/contributing/rule-changes.md
@@ -1,0 +1,30 @@
+---
+title: Rule Changes
+layout: doc
+---
+# Rule Changes
+
+Occasionally, a core ESLint rule needs to be changed. This is not necessarily a bug, but rather, an enhancement that makes a rule more configurable. In those situations, we will consider making changes to rules.
+
+## Proposing a Rule Change
+
+If you want to propose a rule change, [create an issue](https://github.com/eslint/eslint/issues/new?body=**What%20version%20of%20ESLint%20are%20you%20using%3F**%0A%0A**What%20rule%20do%20you%20want%20to%20change%3F**%0A%0A**What%20code%20should%20be%20flagged%20as%20incorrect%20with%20this%20change%3F**%0A%0A**What%20happens%20when%20the%20rule%20is%20applied%20to%20this%20code%20now%3F**%0A%0A%0A) be sure to include:
+
+1. The version of ESLint you are using
+2. The rule you want to change
+3. The code you want to be flagged as incorrect
+4. What happens when the rule is applied to the code without your change
+
+We need all of this information in order to determine whether or not the change is a good candidate for inclusion.
+
+## Accepting a Rule Change
+
+In order for a rule change to be accepted into ESLint, it must:
+
+1. Adhere to the [Core Rule Guidelines](new-rules#core-rule-guidelines)
+1. Have an ESLint team member champion the change
+1. Be important enough that rule is deemed incomplete without this change
+
+## Implementation is Your Responsibility
+
+The ESLint team doesn't implement rule changes that are suggested by users because we have a limited number of people and need to focus on the overall roadmap. Once a rule change is accepted, you are responsible for implementing and documenting it. You may, alternately, recruit another person to help you. The ESLint team member who championed the rule is your resource to help guide you through the rest of this process.
--- a/docs/maintainer-guide/issues.md
+++ b/docs/maintainer-guide/issues.md
@@ -48,12 +48,13 @@ The steps for triaging an issue are:
 1. Once it's clear what type of issue it is, make sure all of the relevant information is provided:
   * **Bugs**: See [bug reporting guidelines](/docs/developer-guide/contributing/reporting-bugs)
   * **New Rules:** See [rule proposal guidelines](/docs/developer-guide/contributing/new-rules)
+  * **Rule Changes:** See [rule change proposal guidelines](/docs/developer-guide/contributing/rule-changes)
   * **Other Changes:** See [change proposal guidelines](http://eslint.org/docs/developer-guide/contributing/changes)
 1. Next steps:
   * **Questions:** answer the question and close the issue when the conversation is over.
   * **Bugs:** if you can verify the bug, add the "accepted" label and ask if they would like to submit a pull request.
-  * **New Rules:** if you are willing to champion the rule (meaning you believe it should be included in ESLint core), then add a comment saying you will champion the issue and add an "accepted" label. As the champion, it's your job to guide the rule creation process until it's complete, so only champion a rule that you have time to implement or help another contributor implement.
-  * **Rule Changes:** if you are willing to champion the change, and the change would not be breaking (requiring a major version increment), then add a comment saying that you will champion the issue and add an "accepted" label. As with new rules, the champion is expected to guide the change through to completion.
+  * **New Rules:** if you are willing to champion the rule (meaning you believe it should be included in ESLint core and you will take ownership of the process for including it), then add a comment saying you will champion the issue. As the champion, it's your job to gain consensus from the ESLint team on inclusion, guide the rule creation process until it's complete, so only champion a rule that you have time to implement or help another contributor implement. Once consensus has been reached on inclusion, add the "accepted" and "help wanted" labels.
+  * **Rule Changes:** if you are willing to champion the change, and the change would not be breaking (requiring a major version increment), then add a comment saying that you will champion the issue. As with new rules, the champion is expected to gain consensus from the ESLint team on inclusion and to guide the change through to completion. Once consensus has been reached on inclusion, add the "accepted" and "help wanted" labels.
   * **Breaking Changes:** if you suspect or can verify that a change would be breaking, label it as "Breaking".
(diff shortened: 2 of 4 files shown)


ilyavolodin:
YWIYBI might be a bit more agressive term then I would've liked. But other than that, LGTM.

btmills:
It sounds like the goal here is to suggest "By default, new rules belong in plugins, but sometimes, if you write something almost everyone will use, we'll make an exception and bring it into core." I agree with @ilyavolodin - it comes off as saying the default state for a rule being rejected. I know it's just semantics because users are always free to write a plugin, but I think the messaging could be more encouraging. (In the spirit of you want it you build it, if you'd like, I can take a stab at some edits to this at some point over the weekend.)

pedrottimark:
@btmills Here are 2 thoughts that came to mind if you can rewrite:
* The difference in my follow-through if I *want* it or if I *need* it.
* The 200 rules in ESLint remind me of the 100 markers in Multi-Dictionary Formatter. The feedback from the front lines was consistent: There are too many 😱 I need another one 😠

pedrottimark:
Thinking at a different level than is usual for my Pooh Bear brain, I wonder if any of us have contacts who could become **early adopters** of this new program:

* companies which are already active contributors to open-source software that they use
* JS meetup groups which already have hack activities
* student groups or inclusiveness organizations

A flurry of (positive) tweets and a few blog articles about the experience could give it some news value beyond our “press release.” So ordinary devs might say, I could do that and want to be a part of that.

Like the momentum behind first-time contributions from tweets by Kent C. Dodds (and others) and Healthy Open Source by Mikeal Rogers.

alberto:
Not exactly thrilled about this, since my contributions are mostly in the rules space, but I understand the motivation.

BYK:
How about simply encouraging people to create a plugin, share and test it out in the wild and come back with data and experience and say: this looks like a good rule to have in core because many people use it now and it is not something-specific.

There can of course be exceptions but they are, well, exceptions.

Now I know this is not really different from what you are saying but the tone as others mentioned will affect the perception and friendliness of the project. Encouraging to write the rule as a plugin and testing out is a "nicer" thing than saying "We don't want this unless really necessary".

alberto:
If we want to encourage the creation and usage of custom rules we should probably do something to increase their discoverability, maybe something like the [grunt plugins listing](http://gruntjs.com/plugins)?

xjamundx:
I'm all for this change and will try to dedicate time to write a blog post
or a screencast about writing custom rules. This is long overdue IMO
because it's the easiest way to scale the project if we can help out with
discoverability.

nzakas:
@ilyavolodin I'm open to other wording, I just want to make it clear that we aren't going to be a rule factory anymore. 

@btmills @byk that's not quite what this is saying. I don't want a stream of "please consider my plugin rules for inclusion." The message here is if your rule is so important that we agree it should be included, then you need to build it. If a rule exists in a plugin, it should stay there. 

@alberto you can still work on rules or enhancements if you want, I just want to shift the discussion so people know that they are expected to implement their own suggestions. If something is accepted and you want to work on it, you still can if the submitter can't fit some reason (you can also submit your own).

Discoverability is something we need to work on. It's a pretty difficult problem, as we don't even have good discoverability for core rules today. I'd say we have to solve it for core rules before we can have a chance at solving it for plugin rules.

So to be clear: the goal here is to **not** accept rules into the core and encourage people to write plugins instead. The case where we accept a new core rule should be extremely rare. According an enhancement to a rule will be more common, but we still want to encourage submitters to make the change themselves. 

We will still address rule bugs that submitters don't fix, and we can still implement truly important rule enhancements.

BYK:
I'm not sure if this is gonna yield the results we are hoping that it would but I also don't think it is bad/dangerous to try it so I'm :+1:.

mysticatea:
:+1: No problem as I still can write rules if I want to get it.

nzakas:
Okay, I've updated the text to try to be a bit nicer. :) Please take a moment to review.

docs/developer-guide/contributing/rule-changes.md
```diff
@@ -0,0 +1,30 @@
...
+title: Rule Changes
+layout: doc
+---
+# Rule Changes
+
+Occasionally, a core ESLint rule needs to be changed. This is not necessarily a bug, but rather, an enhancement that makes a rule more configurable. In those situations, we will consider making changes to rules.
+
+## Proposing a Rule Change
+
+If you want to propose a new rule, [create an issue](https://github.com/eslint/eslint/issues/new?body=**What%20version%20of%20ESLint%20are%20you%20using%3F**%0A%0A**What%20rule%20do%20you%20want%20to%20change%3F**%0A%0A**What%20code%20should%20be%20flagged%20as%20incorrect%20with%20this%20change%3F**%0A%0A**What%20happens%20when%20the%20rule%20is%20applied%20to%20this%20code%20now%3F**%0A%0A%0A) be sure to include:
```
IanVS: I think this should be `If you want to propose a change to a rule…`

IanVS:
Looks good to me.

docs/developer-guide/contributing/rule-changes.md
nzakas: Oops, copy-paste error.

ilyavolodin:
LGTM

kaicataldo:
This looks good to me :+1:

kaicataldo:
I like the friendlier tone :)

docs/maintainer-guide/issues.md
```diff
@@ -85,3 +86,7 @@ Any team member may close an issue if any of the following is true:
 1. The request is outside the scope of the project
 
 In general, it's better to close issues sooner rather than later. It gives people better feedback about the issue affects the project roadmap.
+
+## Consensus
+
+Consensus is reached on issues when there are at least three team members who believe the change is a good idea and no one who believes the change is a bad idea. If consensus cannot be reached, then a consensus among reviewers must be achieved where no one believes it is a bad idea. If consensus among reviewers cannot be achieved, the reviewers will vote and majority wins.
```
BYK: :+1:

BYK:
This latest version looks awesome! :tada:

docs/maintainer-guide/issues.md
```diff
@@ -48,12 +48,13 @@ The steps for triaging an issue are:
...
-  * **New Rules:** if you are willing to champion the rule (meaning you believe it should be included in ESLint core), then add a comment saying you will champion the issue and add an "accepted" label. As the champion, it's your job to guide the rule creation process until it's complete, so only champion a rule that you have time to implement or help another contributor implement.
-  * **Rule Changes:** if you are willing to champion the change, and the change would not be breaking (requiring a major version increment), then add a comment saying that you will champion the issue and add an "accepted" label. As with new rules, the champion is expected to guide the change through to completion.
+  * **New Rules:** if you are willing to champion the rule (meaning you believe it should be included in ESLint core), then add a comment saying you will champion the issue. As the champion, it's your job to gain consensus from the ESLint team on inclusion, guide the rule creation process until it's complete, so only champion a rule that you have time to implement or help another contributor implement. Once consensus has been reached on inclusion, add the "accepted" and "help wanted" labels.
```
alberto: I think there may be times where a contributor may think that a rule should be included but he doesn't have the skills or time to champion it.

nzakas: That's part of the point. If no one is willing to put time into the rule, then that's a good sign that we shouldn't worry too much about it.

alberto: I understand that, but the explanation in parens doesn't reflect what championing a rule really is.

I also think it is interesting for the would-be champion to know the opinion (positive or negative) of other team members beforehand

nzakas: > I understand that, but the explanation in parens doesn't reflect what championing a rule really is.

How would you change the text?

> I also think it is interesting for the would-be champion to know the opinion (positive or negative) of other team members beforehand

That's the part about gaining consensus.

alberto: > How would you change the text?

(meaning you'll be responsible for moving it forward, ensuring it gets done if it finally gets accepted)

> That's the part about gaining consensus.

I understood from the text that the champion would gather consensus **after** he proposed himself.

nzakas: > I understood from the text that the champion would gather consensus after he proposed himself.

Yup, that's exactly right. You say you'd like to champion the rule, then you start discussions with the rest of the team to see if you can get consensus.

I'll update the other wording.

nzakas:
I've updated the text based the last bit of feedback that @alberto have. I'm merging now, as I don't think we should hold this up for other minor issues. We can deal with confusion or other changes as necessary and as they come up.

PR #229, opened 2016-03-24
Docs: Style prerequisite to display wrench icon in fixable rule pages

pedrottimark:
This style rule displays a wrench icon in the left margin of the page for a fixable rule.

It is a prerequisite to the following change from/to in 30 fixable rule pages:

**Fixable:** This rule is automatically fixable using the `--fix` flag on the command line.
(fixable)This rule is automatically fixable using the `--fix` flag on the command line.

We will also move that paragraph immediately after main heading.

The following comments have pictures of the intended layout. The pictures include other future changes to show how each small step moves toward a bigger goal.

P.S. It sounds like a game of Clue: ESLint with the wrench in the margin.

### DIFF
--- a/styles/overrides.css
+++ b/styles/overrides.css
@@ -35,3 +35,7 @@ h6:target:before {
   margin: 0 2px;
 }
 
+/* wrench icon at the beginning of the first paragraph in the doc for a fixable rule */
+h1 + p > span.glyphicon:first-child {
+  margin-right: 1em;
+}


pedrottimark:
On a laptop:

no-multi-spaces fixable-laptop

pedrottimark:
On a tablet in landscape orientation:

no-multi-spaces fixable-tablet-landscape

pedrottimark:
On a tablet in portrait orientation:

no-multi-spaces fixable-tablet-portrait

nzakas:
Hmmm, I'm not a fan of having it stick out of the margin like that - it seems very out of place and looks a bit broken. What if the wrench was at the end of the top heading? Or else, a colored box (yellow?) containing the icon and the text?

pedrottimark:
Thanks for lending your eyes. By the time I make a picture, my “first impression” has been used up.

To explore the solution space,  let me mention a goal is to put the information in a predictable place:
* First priority so you can see at a glance without scanning or scrolling that a rule is *not* fixable.
* Second priority for intentional “banner blindness” so you can ignore it if you don’t care.

A constraint is previous feedback preferring an icon with visible text (which the rule pages currently have, anyway) instead of reducing to an icon with a screen tip.

A quick attempt to align the icon with the margin and therefore somewhat indent the sentence looked mediocre in its vertical alignment, so I’ll work on it some more, and then make a picture.

pedrottimark:
What do you think of a wrench icon aligned **at** the left margin of the page?

On a laptop:
no-multi-spaces fixable1-laptop

On a tablet in portrait orientation:
no-multi-spaces fixable1-tablet-portrait

IanVS:
Personally, I like it aligned as in the most recent screenshots the best.

ilyavolodin:
Interesting. I guess I have a strange taste, but I like the first version a lot more. But I'm flexible, and will not complain if we decide to go with the most recent screenshot.

pedrottimark:
Here is a question about the thought that maybe in the future we can also have "fix" (as an error severity) in https://github.com/eslint/eslint/issues/3626#issuecomment-176833962

Because adding more information to the sentence would make it wrap in most screen resolutions, and the second proposal looks less than ideal for multiple lines, I am interested to know how likely does that change seem?

Of course, a solution is to say less and link to a page for more information.

P.S. A specialized reaction about horizontal alignment would be thumbs left/right instead of up/down.

pedrottimark:
How about a sentence which includes a link and is more consistent with the page about the CLI?

(fixable)The `--fix` option on the [command line](http://eslint.org/docs/user-guide/command-line-interface#fix) automatically fixes problems for this rule.

**Important**: The relative link in the .md file would be (../user-guide/command-line-interface#fix)

On a laptop with possible future limit on line length:
no-multi-spaces fixable1-reworded

nzakas:
Just a heads up: on my tablet, icons in the margin are cut off:
screenshot_2016-03-30-08-08-17

pedrottimark:
@nzakas Uh oh. Can you give the resolution or model of your tablet, so I can analyze the rules for padding and media queries for width of the content container more carefully?

This is the first time I used device mode in Chrome developer tools. Looks like I got a lot to learn.

nzakas:
This is a Nexus 7 using Chrome. I'm not sure of the resolution.

pedrottimark:
@nzakas Thank you. According to http://mydevice.io/devices/ and confirmed by your screen picture, Asus Nexus 7 (v2) resolution in portrait for CSS is 600px by 960px.

My notes didn’t include the gutter for width less than the minimum media query. If you don’t write it down, it didn’t happen!

A way forward is to omit the glyphicons with a **media query** for narrow width.

Which of the following current breakpoints **do you think** are acceptable or unacceptable?

width >= 768px provides 3px to 4px whitespace between glyphicon and left edge
width >= 992px provides 5px to 8px
width >= 1200px provides at least 9px

pedrottimark:
Have a tentative update to the styles for this issue and also to supersede #230

I need to review and test more carefully on Thursday before PR. Lesson learned.

PR #230, opened 2016-03-30
Fix: code block highlight

gyandeeps:
changed the css selectors from
`p.incorrect+div>pre`
to
`p.incorrect~div.highlighter-rouge>pre`

This way it will pick up all the adjacent divs (`~`) with a classname of `highlighter-rouge` and then a apply the css to the immediate `pre` block.

issue: https://gitter.im/eslint/eslint?at=56fbf277e4a8384a1bbd499e


pedrottimark:
Based on previous prototyping, the adjacent selector has a limitation that I was not able to see a way around in a rule page with several sets of incorrect/correct pairs. If the selectors are:

p.incorrect~div.highlighter-rouge>pre
p.correct~div.highlighter-rouge>pre

then any code well after a correct example sentence gets correct style. Imagine this HTML:

```
<p class="incorrect">
<div class="highlighter-rouge"><pre class="highlight">
<p class="correct">
<div class="highlighter-rouge"><pre class="highlight">

<p class="incorrect">
<div class="highlighter-rouge"><pre class="highlight">
<p class="correct">
<div class="highlighter-rouge"><pre class="highlight">
```

There does not seem to be a need for separate code wells in no-constant-condition, but there are a few rules which will need to have separate adjacent examples.

@gyandeeps If you agree with this analysis, would you be willing to adjust this PR to provide explicit adjacent sibling selectors to support up to 3 adjacent code wells?

I will search for any rule docs which have needlessly separated code wells like no-constant-condition and submit a pull request at eslint to merge them.

gyandeeps:
I did not change the html structure, that why i designed my solution like this. I dont want to put a number of 3 out there as it feels so magical.
Based on the current static structure, this css would work well but i am open for more solutions.

pedrottimark:
What I mean is when I tried this, p.correct~div.highlighter-rouge>pre matches all of the code wells which follow, even if they are incorrect. The style rules have equal specificity and they do not according to the closest p to the div.

ilyavolodin:
I think @pedrottimark is correct in this case. I played with styles for a bit, and adjacent siblings selector should be the same specificity for both correct and incorrect styles, so the results might be unpredictable.

pedrottimark:
Agree that a magic number would only be a short-term solution. Another reason we need a long-term solution is that the current replacement rule depends on visible text, so will break in translations.

A long-term solution probably adds distinct content preceding each code fence (by the way, kramdown requires a blank line between content and beginning of code fence) in the md docs which will be replaced entirely.

While we are thinking exactly what that will look like, ideally we want something that can also be the signal to lint correct examples https://github.com/eslint/eslint/issues/2271 as suggested in https://github.com/eslint/eslint.github.io/issues/108#issuecomment-128002598

gyandeeps:
@pedrottimark can u plz fix this issue with #229 ?
If yes, then i can close this PR.

ilyavolodin:
@gyandeeps I'm a bit lost, #229 is not really connected to fenced code highlighting issue. Or am I missing something?

pedrottimark:
@gyandeeps Yes, I am glad to make a new PR with short-term style for multiple adjacent code wells we have been discussing.

@ilyavolodin Discussion in #229 today about cut-off icons somewhat overlaps, so I will probably close that one when it reaches a decision, and start fresh. Do you mind if I remove the newly-added parts of the main.css and work in overrides.css, so I can edit more confidently and you can review the changes more easily?

ilyavolodin:
@pedrottimark No, don't mind at all. I sort of forgot about overrides.css:-) It was painful enough to edit main.css yesterday, that I even contemplated the idea of taking Bootstrap and reskinning it to match our design for a few minutes, but didn't have time to actually do that.

pedrottimark:
Have a tentative update to the styles. I need to review and test more carefully on Thursday before PR.

PR #231, opened 2016-03-31
Fix: Format of multiple code examples and omit icons in narrow margin

pedrottimark:
We will not make a habit of doing 3 things at once:

1. Update #226 and supersede #230 to support up to 5 **adjacent code wells** as a temporary solution especially for no-constant-condition. When we have finished editing rule example sentences and code, only a few rules will need up to 3 adjacent code wells.
2. Supersede #229 to use media query (min-width: 768px) to omit glyphicons in 15px **narrow margin**.
3. Add replace rule which adds a class to display a wrench icon at the left of reworded sentence moved to follow the heading of **fixable rules**: (fixable) The `--fix` option on the command line automatically fixes problems reported by this rule.


pedrottimark:
As an example of issue that wakes me up at night: concerning item 3 about the replacement rule, I though about the yeoman setup which provides a starter rule doc page, and realized my ignorance about the environment in which doc pages for **custom rules** are processed and displayed.

I was thinking without the CSS rule, the wrench icon would at least appear inline with the paragraph, but that assumes Jekyll and Liquid templates run to replace the glyphicon. What can we assume?

Is it more robust **progressive enhancement** for markdown to be **Fixable:** The `--fix` option …?

Therefore replace `<p><strong>Fixable:</strong> ` with `<p class="fixable">`

I have mixed feelings about matching **language text** in replacement rules, especially waking up to see the cn site and jp is coming. If the replacement rule had to match **Fixable:** instead of the language-independent `(fixable)` notation, then I guess each language repo could add replacement rules to match the translated text.

nzakas:
We can't assume anything about custom rules. We encourage people by providing a documentation template, but those don't make it onto our site, so we don't need to worry about that.

I'd say don't worry too much about multiple language support right now. I think we have a lot of work to do to coordinate with cn.ESLint.org, so I don't won't to get too hung up on that right now. Let's make the English site as awesome as possible first and then we can figure out what, if anything, we want to change for other languages.

pedrottimark:
Because no-constant-condition now has only one code well as part of ESLint 2.6.0, I reduced support to **3 adjacent code wells** which is the maximum required (for vars-on-top).

I am not a fan of “magic numbers” but this method seems okay for a while.

Some rules have more than 3 adjacent code wells following an old inconsistent example sentence. They will be simplified as part of the remaining **distinguish examples** process.

pedrottimark:
Looks like CSS rule selectors on the **cn** site might just require `p + p` for the en + cn pattern:
`p.incorrect + p + div > pre` and so on
`p.correct + p + div > pre` and so on
`p.fixable + p:before` and `p.fixable + p`

pedrottimark:
After these prerequisite changes, I will submit a pull request to eslint for the 30 fixable rule docs:

(fixable) The --fix option on the command line automatically fixes problems reported by this rule.

ilyavolodin:
@pedrottimark Thanks for reminder!
Unrelated: Anyone knows how to turn off gmail integration with Github? As soon as read an email it's automatically removed from my notifications in github, and I keep forgetting about things outside of ESLint core repository.

PR #232, opened 2016-04-03
Fix: /docs/developer-guide/code-path-analysis is pointing to the assets directory

kaicataldo:
http://eslint.org/docs/developer-guide/code-path-analysis/ is currently inaccessible. This seems to be because we have /docs/developer-guide/code-path-analysis.md and the assets directory docs/developer-guide/code-path-analysis/, and GitHub Pages is trying to serve the directory instead of the HTML file.

http://eslint.org/docs/developer-guide/code-path-analysis.html works because it is explicitly requesting the HTML file rather than the directory. Easy fix is to move /docs/developer-guide/code-path-analysis.md to /docs/developer-guide/code-path-analysis/index.md and change the links to the assets.

Alternatively, we could just rename the assets folder to something like code-path-analysis-assets, but I personally think the index.md solution makes more sense, given that the rest of the site works that way as well.

### DIFF
--- a/docs/developer-guide/code-path-analysis/index.md
+++ b/docs/developer-guide/code-path-analysis/index.md
@@ -20,7 +20,7 @@ if (a && b) {
 bar();
 ```
 
-![Code Path Example](./code-path-analysis/helo.svg)
+![Code Path Example](./helo.svg)
 
 ## Objects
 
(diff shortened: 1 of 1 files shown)


mysticatea:
Thank you!
I had not noticed it.

kaicataldo:
I should have pinged the team :)

ilyavolodin:
LGTM. Thanks.

Issue #186, opened 2016-01-27
Improve information architecture of site


nzakas:
The information architecture we have at present is mostly an accident of how our folders were setup initially. There was only a small amount of thought put into how best to organize the documentation, as we just wanted to write as much documentation as possible without worrying about the best way to present it.

Now that we have a massive amount of documentation, we could really use some help figuring out better ways of organizing it. Any and all recommendations welcome! (Be gentle, the team doesn't have any information architecture experience or knowledge.)

ilyavolodin:
Some of the things to look into would be to break up large articles like http://eslint.org/docs/developer-guide/working-with-rules into smaller pages, or provide internal navigation through them. Make configuration options more prominent and easier to find.
Also, keep in mind that majority of our documentation is auto-generated from .md files, so we don't always have full control over layout and structure of the output.

pedrottimark:
Example of a simple inconsistency which pushes me out of the flow a bit each time I learn a new rule:

http://eslint.org/docs/rules/ has rule string as hyperlink text followed by hyphen and rule description in all lowercase, for example:
comma-dangle - disallow or enforce trailing commas

http://eslint.org/docs/rules/comma-dangle heading has rule description in title case followed by rule string in parentheses, for example:
Disallow or Enforce Dangling Commas (comma-dangle)

nzakas:
@pedrottimark and what would be the solution?

pedrottimark:
@nzakas Hello! Here are 2 example of tactical goals and possible changes to achieve them:

* Make it easier to scan the lists in http://eslint.org/docs/rules/

a. Move icons (for recommended and fix) from the end to beginning of line, possibly as columns of tables (without borders) instead of the bulleted lists.

b. Slightly reduce punctuation, possibly use some white space separation, but not separate columns.
comma-dangle: disallow or allow trailing comma

* Make a clearer visual connection between cause and effect when you click a rule link, and then see heading on destination page.

a. Heading of rule page has same text as line in rule list, but different style, of course.

b. If the icons are important enough on the rule page (pending analysis, see below), might hang them in the margin at the left of the heading. Provide screen tip via title attribute. If there is enough supplementary information, the icons are hyperlinks to pages about recommended and fix.

Does that help? What do you think about the problem-solving approach? And the specifics, of course?

Can you compare and contrast your vision as owner to the following possible way forward: a team of people (who respond to the call) review interior content, evaluate according to heuristics, and discuss improvements using a consensus of approach.

[...]

PR #234, opened 2016-04-05
Docs: Style rules for columns of icons in tables of rules

pedrottimark:
See https://github.com/eslint/eslint.github.io/issues/186#issuecomment-196460629

### DIFF
--- a/styles/overrides.css
+++ b/styles/overrides.css
@@ -35,6 +35,25 @@ h6:target:before {
   margin: 0 2px;
 }
 
+td .glyphicon {
+  margin: 0; /* within table in rules page */
+}
+
+th,
+td {
+  padding-left: 0.5em;
+  min-width: 1em; /* width of glyphicon for a column of empty cells */
+}
+
+th:first-child,
+td:first-child {
+  padding-left: 0;
+}
+
+tr {
+  vertical-align: baseline;
+}
+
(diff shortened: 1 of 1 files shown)


pedrottimark:
Added th selectors for cell padding rules, in case any tables need headings.

pedrottimark:
Added baseline alignment because table cells can wrap to multiple lines on narrow screens.

pedrottimark:
Thinking ahead to the next step, how about a separate **Replaced by** column under **Removed**?

1280px wide on laptop:
removed laptop 1280


768px wide in portrait orientation on iPad 2:
removed ipad 768


600px wide in portrait orientation on Nexus 7:
removed nexus_7 600


320px wide in portrait orientation on iPhone 5:
removed iphone_5 320

nzakas:
Do we need the description for the removed rules? It would clean things up a lot if not.

pedrottimark:
For sure. And it makes that table look very different from the others. A little bit different is dangerous for clear technical communication.

Even if the index page omits descriptions, it looks like removed rules would appear in search results (with heading text from their doc).

While we’re at it, omit the columns for check mark and wrench, too?

600px wide in portrait orientation on Nexus 7:
removed2 nexus_7 600


320px wide in portrait orientation on iPhone 5:
removed2 iphone_5 320

PR #235, opened 2016-04-11
WIP: Initial check for demo rewrite

ilyavolodin:
This is quick starting point for possible approach to demo re-write. It uses React with JSX, but ES5 only. JSX is transformed on the fly by require plugin. No build steps are necessary. I created a wrapper for orion and component to display errors, but haven't gotten around to finishing configuration (even in the state that it is now in the demo). Would really appreciate feedback from people who are familiar with react (specifically @btmills and @xjamundx), since this is the first time I wrote anything in React.

P.S. I also modified demo/index.html to be a full page for development ease. It should be converted back to be partial page once the rewrite is done (if we are going to stick with this approach).


js/app/demo/message.jsx
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+define(['react'], function(React) {
+    return React.createClass({
+        displayName: 'Message',
+        render: function() {
+            return (
+                <div className={"alert alert-" + (this.props.value.fatal || this.props.value.severity === 2 ? "danger" : "warning" )} title={this.props.value.message}>
+                    {this.props.value.line}:{this.props.value.column} - {this.props.value.message} (<a href="http://eslint.org/docs/rules/{this.props.value.ruleId}">{this.props.value.ruleId}</a>)
```
btmills: You'll need to do string concatenation in the link:

```jsx
<a href={"http://eslint.org/docs/rules/" + this.props.value.ruleId}>...</a>
```

js/app/demo/messages.jsx
```diff
@@ -0,0 +1,26 @@
...
+                    <div id="results">
+                        <div className="alert alert-success">Lint-free!</div>
+                    </div>
+                )
+            } else {
+                return (
+                    <div id="results">
+                        {
+                            this.props.values.map(function(message, index) {
+                                return <Message key={index} value={message} />;
```
btmills: `key`s are hard in React. For a long time, I used `index` too, which silenced the warning, but didn't actually solve the problem. `key` should be unique (and constant) across renders, so perhaps something like the concatenation of line, column, and rule ID? I'll try to explain via example.

1. Paste the following code in the demo editor:

    ```js
    alert(1);
    alert(2);
    ```

2. You'll see two messages, as expected:

    screenshot 2016-04-11 11 31 34

3. Right click -> Inspect on the rule link in the second message. In the dev tools, the `<a>` element should be highlighted.

4. Comment out the first line:

    ```js
    alert(1);
    alert(2);
    ```

5. The first message goes away, and the second *appears* to remain, but the `<a>` is no longer selected in the element inspector. Because the `key` changed, React had to re-render the entire DOM for messages. Ideally, React would be able to detect that the second message is unchanged and wouldn't have to re-render it.

6. Try giving the messages a unique `key`:

    ```diff
    diff --git a/js/app/demo/messages.jsx b/js/app/demo/messages.jsx
    index 40a78fa..ee8b552 100644
    --- a/js/app/demo/messages.jsx
    +++ b/js/app/demo/messages.jsx
    @@ -14,8 +14,10 @@ define(['react', 'jsx!message'], function(React, Message) {
                     return (
                         <div id="results">
                             {
    -                            this.props.values.map(function(message, index) {
    -                                return <Message key={index} value={message} />;
    +                            this.props.values.map(function(message) {
    +                                var key = '' + message.line + ':' + message.column + ':' + message.ruleId;
    +
    +                                return <Message key={key} value={message} />;
                                 })
                             }
                         </div>
    ```

7. Repeat steps 1-4, but this time, rather than re-rendering every message, React only had to remove the first message's DOM element. The `<a>` remained selected. In fact, you can uncomment the first line again, and React will insert a DOM element at the beginning of the message list, but it won't touch the DOM for the message on the second line.

btmills: Typically, I'd prefer to pass `line`, `column`, `ruleId`, and `message` as props directly on `Message`. This makes it slightly more convenient to access props within the component (don't need `.value`) and easier to validate via `PropTypes` (don't need to nest within `PropTypes.shape`). With a transpiler, I'd do that with `<Message key={key} {...message} />`. Without a transpiler, if `Object.assign` had slightly higher [compatibility](http://kangax.github.io/compat-table/es6/#test-Object_static_methods_Object.assign) or we had a polyfill, I'd do it with `React.createElement(Message, Object.assign({ key: key }, message))`. Since we have neither of those right now, I'm leaving a note and saying this is the best way to do it unless or until one of the above changes.

btmills:
Architecture looks good so far. :+1: I had a couple comments on style I've picked up and some React gotchas, but nothing significant. I didn't dig into configuration yet. Have you thought about how you'll want to handle that? I assume [controlled components](https://facebook.github.io/react/docs/forms.html#controlled-components) will be the way to go.

ilyavolodin:
@btmills Yes, I was planning on controlled components (and even started going down that path), but I got lost in `this.props` vs `this.state` and even though state was updated, it wasn't communicating it up the stack. I only spent a few minutes on it, though. So I'm sure it's fixable, just ran out of time. Thanks for the pointers. I thought about unique keys, but just ended up using `index` for now.

ilyavolodin:
Updated PR. The new demo now has feature parity with the current one (except it actually works). @btmills @xjamundx could you provide some feedback? I noticed that checking/unchecking checkboxes has a very noticeable delay. I'm not really sure why that is. Also I'm really not sure I'm dealing with events correctly. Any feedback would be appreciated.

xjamundx:
Sure, let me pull it down and take a look

kaicataldo:
Nice work, @ilyavolodin!

Spent some time familiarizing myself with the code tonight. Any thoughts on how we can pull the rule options in? Happy to work on the rule options and stuff if I can figure out how to get them into the state. 

Also seeing the delay on my end - I wonder if it's re-rendering every rule on each click...If that's the case, we maybe want to fix that before adding even more with the rule options :)

One last thing - Jekyll takes a while to update each time I save a JS file...Is that normal?

ilyavolodin:
@kaicataldo I was looking at this react component before: https://github.com/mozilla-services/react-jsonschema-form but, unfortunately, we can't use it, since it's all ES6. But something along those lines would work. We would need to export schemas from ESLint rules, but that's fine, and can be done as part of the release script (in the future, we would want to export the whole metadata anyways).

Jekyll is really slow on updates in the latest version, that's why I turned demo page from partial into a full page. I just run simple http-server instead of Jekyll while developing, but you can be fancy and run something with hot-reload as well.

P.S. @xjamundx any feedback?

xjamundx:
Looking at it again now.

xjamundx:
Okay so first impression is that I'm really impressed!

js/app/demo/app.jsx
```diff
@@ -0,0 +1,88 @@
...
+            };
+        },
+
+        handleChange: function(event) {
+            this.setState({ text: event.value }, function() {
+                this.lint();
+            });
+        },
+        updateOptions: function(options) {
+            this.setState({options: options}, function() {
```
xjamundx: In general let's avoid the `setState` callback function wherever possibly. It can lead to state bugs and may be causing some of the slowness we're feeling.

What if we just ran the linting on a timer? Every second we run the linting. It doesn't care about events or anything at all. My hunch is that that would help with the performance of everything. We could even have "hasChanged" checks to make sure it's not doing extra work, but I bet it would be way faster, because here we have setState being called inside of setState callbacks and it's probably  causing some silliness

js/app/demo/rulesConfig.jsx
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+define(['react'], function(React) {
+    return React.createClass({
+        displayName: 'RulesConfig',
+        getInitialState: function() {
+            return this.props.config;
+        },
+        buildRules: function(state) {
```
xjamundx: hit: would be more canonically called `this.renderRules()` and it should already have access to the state, so may not be needed to pass it in......

xjamundx:
Overall structure-wise the components themselves seen fine, but it's already complex enough where you can see the benefit of something like redux/FLUX. But that's another 4 or 5 files of boilerplate that might be a pain, so your mileage may very. 

With a redux architecture you'd get rid of all of the `setState` calls, everything would be props and instead of any explicit event emitting you'd basically emit actions, which would make changes to the global state. Overall it simplifies the managing of state, because you only manage it one place instead of throughout several different components and in general it gets rid of all of the callbacks you see here.

Of course if we can get the current system perofmantly to work as-is, let's do that first :)

Overall 👍

btmills:
@xjamundx I'm glad you brought up Redux. I'd been struggling to figure out the right state shape to make this work performanly, but I kept coming to the conclusion that external state would be easier.

I'd like to avoid adding in another library if possible (even though Redux is tiny), but the `setState` calls reveal how complicated the state is like this.

js/app/demo/app.jsx
```diff
@@ -0,0 +1,88 @@
...
+                this.lint();
+            });
+        },
+        lint: function() {
+            var results = eslint.verify(this.state.text, this.state.options);
```
xjamundx: if you wrap this whole thing like this it seems pretty fast:
```
setTimeout(function() {
                var results = eslint.verify(this.state.text, this.state.options);
                this.setState({
                    messages: results,
                    name: this.state.text
                })
}.bind(this), 0)
```

xjamundx: or maybe we could debounce it for similar effect

js/app/demo/app.jsx
ilyavolodin: I'm using callback for `setState` all over the place. I assumed that's the proper way to do this. Pretty much every component under `configuration` has interactive elements, and when they are changed, I update state and then in the callback do `this.props.onUpdate(this.state)`. Is there a better way to kick information to the parent component?

js/app/demo/rulesConfig.jsx
ilyavolodin: Sure. Incomplete refactoring:-) This was originally private function, and I didn't fully refactor it. I have to say, JSX isn't all that much fun to deal with when you have to do complex looping... Simple `map` is fine and comfortable, but if you need to do something more complex in a loop, it sort-of becomes a mess pretty quickly.

js/app/demo/app.jsx
ilyavolodin: Hmm... Interesting. It was actually very fast up until I added rule configuration in. With just parser options and environments it was VERY responsive. Not sure why rules would affect this particular code.

ilyavolodin:
I was thinking about Redux for state management. There are a few problems with bringing Redux in.
1. Another external dependency. Not a huge deal, but still...
2. Since I don't want to have build step, I'm stuck with ES5. And because Redux requires pure functions, without destructuring and spread it would be a LOT of code! I guess I could bring in something like lodash-fp (or whatever is the current flavor of the month), but that's yet another dependency.

I could, obviously, write my own simple external state without using Redux, but that feels like reinventing the wheel...

hzoo:
@ilyavolodin since you are already using requirejs + jsx, why don't you just switch that out for babel then? https://github.com/mikach/requirejs-babel, essentially switching out jsxtransform for babel

xjamundx:
I'm working on a fork of this PR that uses `browserify` + `babelilfy`. If
there's a simple build step, but we keep the compiled version in the repo
(like the parser deom), I don't think it's actually a huge problem IMO. Let
me see if I can get it working....

ilyavolodin:
@hzoo Does Babel team sits down and says "What should we fork today to create a babel version of?":-) Yeah, I guess that would work, although I would imagine performance would suffer a bit, but probably not too much comparing to jsx plugin. @xjamundx That's what I was suggesting originally, but I think @nzakas is against it. Sort of complicates things and if you forgot to run build process before checking in things, you will get out of sync.

xjamundx:
@nzakas loves us and will give in eventually ;-)

xjamundx:
I think I got it working...I'll send up a PR to your PR for further discussion :-p

ilyavolodin:
Nice. Even if I end up not using your version, I will definitely learn something new, and might pick some parts of your implementation anyways.

hzoo:
@ilyavolodin I don't think the babel team made any of the integrations - we just moved all of them under the babel org if that's what you are talking about haha.

xjamundx:
Here you go:
https://github.com/eslint/eslint.github.io/pull/239

nzakas:
I'm still of the mind that if we need a build step, then we should be managing the demo as a separate app and not part of this.

kaicataldo:
@ilyavolodin That React component looks really promising. It looks like they have [a CDN](https://github.com/mozilla-services/react-jsonschema-form#as-a-script-served-from-a-cdn) we could use.

ilyavolodin:
@kaicataldo Unfortunately we can't use it, first because it's ES6 (well, until we either create a full blown build system, or maybe use `requirejs-babel`) and also because it's very much incomplete and requires specific fields in the schema. But something along the lines of that component would work.

hzoo:
@ilyavolodin It being written in es6 shouldn't stop us from using it since it would build out/be consumed as es5?

ilyavolodin:
@hzoo For that we would need a build process. I'm still trying to avoid it for now. I have a feeling that I would need to give in at some point, but I'm afraid that build process (which means we should deploy to our DO server instead of Github) will complicated release process significantly. If Mozilla's component would be complete and working out of the box, I would probably give in, but it's far from complete, it can't handle `enum`, `one-of` and a bunch of other JSON schema constructs that we use in our rules.

xjamundx:
I'll iterate on my built version on Monday. Build before commit seems
pretty straight forward IMO

xjamundx:
A few more perf suggestions coming soon, trying it a few ideas around
minimizing updates.

PR #244, opened 2016-04-18
Docs: Replacement rules for examples of additional code

pedrottimark:
As discussed in https://github.com/eslint/eslint/pull/5807#issuecomment-207621143 and https://github.com/eslint/eslint/pull/5807#issuecomment-207689759, and then added to guidelines in https://github.com/eslint/eslint/issues/5446#issuecomment-191306528, the word additional makes it explicit that code is the same as or related to a preceding example.

This change will add the class which displays icon and background-color in recently edited rule docs:
* allowEmptyCatch option in no-empty
* allowArrowFunctions option in func-style
* exceptions option in id-length

### DIFF
--- a/_layouts/doc.html
+++ b/_layouts/doc.html
@@ -3,7 +3,19 @@
 <main class="doc">
   <article class="container">
     {% include ad.html %}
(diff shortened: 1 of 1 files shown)


ilyavolodin:
LGTM. Thanks.

PR #245, opened 2016-04-18
Docs: Style rules for glyphicons at narrow widths

pedrottimark:
To implement the pictures in https://github.com/eslint/eslint/issues/5774#issuecomment-208400114 when screen width is less than 768px:

* Display glyphicons inline instead of hiding them. That implies change to display incorrect thumbs-down and correct thumbs-up at left of example sentences instead of code wells; see https://github.com/eslint/eslint/issues/5774#issuecomment-208401154
* Allow code elements to wrap, especially because some example sentences now have long options.

### DIFF
--- a/_layouts/doc.html
+++ b/_layouts/doc.html
@@ -2,17 +2,18 @@
 {% include menu.html %}
 <main class="doc">
   <article class="container">
-    {% include ad.html %}
     {{ content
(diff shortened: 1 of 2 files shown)


pedrottimark:
1. Remove ad to avoid a regression!
2. Add `icon` class in replacement rules to dry out style selectors.
3. Add `recommended` and `removed` icons to lean toward https://github.com/eslint/eslint/issues/5774#issuecomment-208396092

nzakas:
Sorry, we lost track on this. Looks good!

PR #252, opened 2016-04-26
Remove ad (fixes #251)

### LINKED ISSUE #251: Remove ad
The jQuery Foundation has requested we remove the ad from the site. We don't really need the extra revenue since they will be footing the bill for most stuff. :)

pedrottimark:
While testing locally, noticed it removes the gap in rule docs which have code near the top.

With ad:
251 with ad

Without ad:
251 without ad


nzakas:
Yup, that was a side effect we had to live with. Looks nicer without the ad. :)

PR #255, opened 2016-04-29
Chore: Remove old unused docs

nzakas:
I'm pretty sure we don't link to these anywhere at this point (duplicates are in `user-guide`).

### DIFF
--- a/docs/command-line-interface/index.html
+++ b/docs/command-line-interface/index.html
@@ -1,6 +1,0 @@
----
-layout: redirect
-sitemap: false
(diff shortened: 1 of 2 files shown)


ilyavolodin:
@nzakas Those are not for the users, they are for search engines. We renamed a few pages some time ago, and best practice is to put a redirect when you do something like that. It's been a while, so maybe it's fine to remove those now, though.

nzakas:
Oh right. Up to you .

ilyavolodin:
Those pages has been removed for a long time, I think it's fine to remove redirects now.

PR #257, opened 2016-05-16
Adds `no-irregal-whitespace` as rule that options were added.

mysticatea:
The following options were added to `no-irregal-whitespace`.

- skipStrings
- skipRegExps
- skipTemplates


alberto:
The correct rule name is `no-irregular-whitespace`

mysticatea:
Ouch!

mysticatea:
I updated it. Thank you, @alberto !

PR #260, opened 2016-05-19
Fix parsing of rule name for indexing

### LINKED ISSUE #258: Website build is failing
Website build (more specifically algolia re-indexing) has been failing for the last 5 builds. That means that our search is 3 version behind (last successful update was for version 2.9.0). Could somebody with some experience with Ruby take a look? I don't feel right bugging @pixelastic every time we have an issue.

cc: @eslint/eslint-team

pixelastic:
Seems like you introduced a new pattern on the way page title are handled. Previously, I was parsing patterns like `rule description (rule name)` to correctly build the hierarchy. The regexp was failing on the new pattern of `rule name: rule description`.

I updated the code to take both patterns into account and fallback to the whole title in case a new pattern is added (instead of failing).

Should fix #258

### DIFF
--- a/_plugins/search.rb
+++ b/_plugins/search.rb
@@ -80,6 +80,34 @@ class EslintCustomSearchHelper
     %w(h1 h2 h3 h4 h5 h6).include?(tag_name)
   end
 
+  def self.extract_rule_name_and_description(rule)
+    # Excluding main page
+    return {} if rule == 'Rules'
+
+    # Page titles can have one of this two patterns
+    # 1. rule-name: description
+    matches = rule.match(/^(.*): (.*)$/)
+    if matches
+      return {
+        name: matches[1],
+        description: matches[2]
+      }
+    end
+
+    # 2. description (rule-name)
+    matches = rule.match(/^(.*) \((.*)\)$/)
+    if matches
+      return {
+        name: matches[2],
+        description: matches[1]
+      }
+    end
+
+    {
+      name: rule
+    }
+  end
+
   # Returns information about the item place in the page hierarchy
   # - main category (eg. Rule, Blog, etc)
   # - subcategory (eg. accessor-pairs, ESLint v1.7.0 released)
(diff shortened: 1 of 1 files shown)


ilyavolodin:
Oh! I didn't even realize we are parsing titles of the documentation pages! Thank you very much for digging into this! Could you please sign jQuery CLA (we are now part of jQuery Foundation, so CLA changed), and I will merge this in.

pedrottimark:
@pixelastic Sorry, did not know to tell you about that change 😳

Soon the gensite process in Makefile.js will use rule metadata, and then all the rule doc pages will have main heading of `rule-id: rule description` for consistency with the rules index page

Here are two more changes coming, in case they are breaking:
* The rules index page will change from lists to tables
* Pages other than individual rule docs might have a table of contents `<ul id="markdown-toc">...</ul>` with links to anchors in any h2 or h3 headings on the page

pixelastic:
@pedrottimark I don't think this will break anything.

The plugin looks at the HTML files generated by Jekyll. It automatically skips the files that are "not interesting" (like index pages). This is actually a whitelist/blacklist code in the `_plugins/search.rb` file. 

It then looks in those files for every `p`, `li`, `tr` and `h1`, `h2` etc. It will just silently ignore all the rest. So as long as the titles are in HTML headings and the content in `p`, `li` or `tr` it should find it just fine :)

I'll read and sign the CLA shortly.

pixelastic:
Apparently I did not sign the CLA correctly (was using my real name and not the alias used in my git config). I signed again, with the correct name. Will the robot catch that or should I re-submit the PR?

ilyavolodin:
It should be able to update automatically.

ilyavolodin:
@kborchers Would you mind helping us with CLA issue here? Looks like name doesn't match: 

```
Pixelastic <tim@pixelastic.com>: Name on signature (Tim Carry) doesn't match name on commit.
```

@pixelastic We'll figure out CLA issue and merge it in. Thanks so much for fixing this!

pixelastic:
You're welcome :)

ilyavolodin:
Awesome! Thanks @kborchers

ilyavolodin:
Build is now passing. Search index is updated, although it seems a bit strange:

image

pixelastic:
It's the "expected" behavior, but I agree it seems too crowded in that example.

We display the rule name in the left column, then the `Rule description > potential sub heading` in the right one. When the rule description is really long, and the heading quite short, it seems like we are repeating the same content.

I'm not sure how we could do that any other way, but all feedbacks are welcome.

ilyavolodin:
I only noticed after I posted that that each of those link to different sub-header on the page. I think that's perfectly fine. I though it was linking to the same thing 4 times, which would be strange.

PR #266, opened 2016-06-16
Docs: increase horizontal padding (fixes #265)

pedrottimark:
* increased [EDIT minimum] padding from 0.5em to 1em = 17px
* increased pre padding from 2px to 4px
* corrected gutter between content and toc to 6em double default left padding and display at > 1024px

PR #269, opened 2016-06-20
Move styles to less

ilyavolodin:
Turned out to be easier then I expected. I didn't have to recreate our theme, since, apparently it was already checked in. I combined all various stylesheets that we had into a single file. Converted them to less and cleaned them up to use bootstrap variables where possible (except for overrides.less, that's just a direct copy of css file, since it will need some major tweaks).
@pedrottimark Could you please verify that everything looks correctly? You need to do `npm install` and `npm run less` before running jekyll.
main.css file is still checked in, so we don't need build step for site deployment. Just run `npm run less` locally, and it will generate a new main.css file that will be checked into github repository.


pedrottimark:
@ilyavolodin Super, thank you for putting the styles on such a strong foundation.

Ouch, how did I miss that bootstrap_variables.less was already in the project?

How do you want me to move forward balancing the urgent with the important?
* submit changes in #266 to correct the regression for large screens first, and verify this second
* or start here first

ilyavolodin:
@pedrottimark Whichever way is easiest for you. I, personally, would probably do both at the same time. Overrides needs some major tweaks, so I might just start with a clean slate and write whatever is necessary in less directly.
For example, all of the style changes that you've done to `container` class should be reverted, since they are the ones that are making page full length (and if that was the goal, then we should've just changed `container` to `container-fluid` instead of overriding styles).

alberto:
Any reason why you used less and not Sass?

I would also leave the comments on the color classes.

ilyavolodin:
@alberto No reason other than when our site was created, there was only less version of Bootstrap, so all of the customizations were done in less.
I personally work with Sass much more now then with Less, however I still prefer Less (but I can live with either, doesn't really matter to me much).
bootstrap_variables.less was not created by me, so can't add comments there. The rest of the files were auto-generated from css files, so, again, can't really add comments to those either.

pedrottimark:
@alberto You mean keep the comments from `rouge.css` into `rouge.less`? Yeah, I needed those comments to understand what was going on there.

Agree with prefer Less, but can live with either preprocessor.

alberto:
I mentioned Sass because I think it's much more popular nowadays (and I like it better :)) and it can be integrated with github pages without a build step, which I thought was an important thing.

@pedrottimark yeah, I was talking about those comments.

ilyavolodin:
@alberto Good point about not needing build step for sass. But we would need to convert our files to sass in that case, and I didn't have time to deal with that (and not remembering jekyll support for sass, didn't see the point either).

I converted our css files to less using automatic conversion tool, which stripped out comments. I guess we could manually add them back.

pedrottimark:
I can manually add comments in files where there is a consensus that the information is important.

For the limited amount of preprocessor features we need and that the automatic conversion used, it looks like there might be a very short list of differences between LESS and SCSS, like variable notation.

nzakas:
This is the reason I still use regular CSS. :)

pedrottimark:
This source says Jekyll 2.4.0 has a Saas converter: http://markdotto.com/2014/09/25/sass-and-jekyll/

pedrottimark:
@alberto @ilyavolodin Bootstrap questions:

How does it decide the width from the breakpoint?

Here is the original CSS:
```
@media (min-width:768px){.container{width:750px}}
@media (min-width:992px){.container{width:970px}}
@media (min-width:1200px){.container{width:1170px}}
```

This seems to be the corresponding LESS:
```
// Media queries breakpoints

@screen-xs:                  480px;
@screen-sm:                  768px;
@screen-md:                  992px;
@screen-lg:                  1200px;

// Grid system
@grid-gutter-width:         30px;
```

I get 1170 = 1200 - 30 because that is breakpoint minus gutter, but what about:
* 750 = 768 - 18
* 970 = 992 - 22
* Interesting no occurrence of 480 in the CSS file. Is there a way to control how many breakpoints?

ilyavolodin:
@pedrottimark Here's where it is: https://github.com/twbs/bootstrap/blob/master/less/variables.less#L342

pedrottimark:
@ilyavolodin Thank you. What do you think of this plan?
1. I give you some screen pictures to review of minimal additional changes to #266 to return to original appearance of home page and limit line length on internal pages within a centered container (main remaining detail work is give the glyphicons enough padding near the breakpoints, which the original design did not do).
2. While you decide whether the centered-but-somewhat-asymmetric appearance is adequate immediate solution, I continue testing it to see whether it is solid enough for ESLint 3 launch.
3. After that, I get experience with LESS by compiling the theoretically-original Bootstrap theme and diffing with original CSS (both transformed to rule-per-line for purpose of comparison) to verify there were no more changes than we can account for.
4. I continue to get experience with LESS by compiling the files you converted and browsing local version of site compared to an earlier version to see if there are any unintended differences.
5. Work with you or other team members to reach a long-term design and implement it in Bootstrap with as few overrides as possible. Main additions from what we have now: Including side-by-side example code on wide screens. Decide what to do about TOC on non-rules pages.
6. Implement the long-term design in SCSS.

ilyavolodin:
OK, I updated this PR with original styling. @pedrottimark I removed most of the code you added with last PR to bring the site back to normal look and feel. I also removed limitation on line-length for now (I left it commented out in overrides.less for the future). Since most of the code you added in last PR is not testable right now (tables for rules page, etc.) I had no way to verify that they are working correctly and not breaking anything, so for now I removed that whole block. Once we get closer to changing rules page layout, we can add them back in.
Since we just release 3.0 I think it's important that we make site workable again, because people will be visiting it a bit more often for the next couple of days. So I think we should merge this in, and bring back the old look and feel, and then do **small** incremental changes to make adjustments. If we want to make more dramatic changes to the site, I think it's worth reaching out to the community and seeing if there are any designers that are willing to work with us.

ilyavolodin:
@eslint/eslint-team Can we merge this in? I'd really like to get website back into reasonable layout.

nzakas:
@ilyavolodin completely up to you. If you're happy, go for it.

ilyavolodin:
Ok, I'm going to merge this in to get website back into a working order. @pedrottimark We can add all of the rest of your enhancements on top of this PR one at a time.

PR #274, opened 2016-07-15
remove io.js

hzoo:
We aren't testing this anymore


hzoo:
Ah I changed to use the noreply email so I was different

PR #276, opened 2016-07-20
Fixing table rendering

ilyavolodin:
Hopefully this will fix issue with extra `</col>` tags being displayed on the page. Can't test it locally, since it worked fine with the previous version


ilyavolodin:
Going to merge this in to verify that it's working correctly

PR #280, opened 2016-08-13
Add striped style and use table style for deprecated/removed rules too

vitorbal:
Added striped table style to all lists of rules in /docs/rules. Also made it so the deprecated & removed rules lists use the same table style as main list of rules above it.
Should look like this now:
screenshot 2016-08-13 02 28 56


docs/rules/index.md
```diff
@@ -45,21 +45,55 @@ The `--fix` option on the [command line](../user-guide/command-line-interface#fi
 
 {{ site.data.rules.deprecated.description }}
 
-<div class="deprecated-rules" markdown="1">
-|Deprecated rule|Replaced by
-|---|---
-{% for rule in site.data.rules.deprecated.rules %}|[{{rule.name}}]({{rule.name}})| {% for replaced in rule.replacedBy %} [{{replaced}}]({{replaced}}) {% endfor %}
+<div class="rule-list deprecated-rules">
+<table class="table table-striped">
+
```
ilyavolodin: Could you add colgroup, so we could make both removed and deprecated tables align?

ilyavolodin:
Now that I look at it, I'm not sure I like striped table all that much, what we have currently is much cleaner, but pretty much as readable. Maybe we should keep current design (for normal rules, we should still update deprecated and removed tables), but add hover style? This way, if you are having trouble tracing columns, you can always mouse over it to highlight it.
@nzakas What do you think?

docs/rules/index.md
vitorbal: Oh, of course!

vitorbal:
@ilyavolodin I agree, I think I also prefer keeping the current design as it is cleaner, and just adding the hover effect. Let's see what @nzakas thinks!

vitorbal:
Updated the commit to add colgroups.

nzakas:
Hover only works if you can use a mouse. I like the striping much better than what we currently have.

vitorbal:
I wouldn't mind going with the striping. @ilyavolodin what do you think?

ilyavolodin:
Well, if @nzakas feels so strongly, I'm not opposed. Let's go with what we have.

ilyavolodin:
Hmm.. for some reason Jekyll page generation failed with non-descriptive error message. I'm not really sure what to do next...

vitorbal:
I pushed an empty commit to force a rebuild and it passed now! 🎉

PR #287, opened 2016-09-14
Docs: add commas between 'replaced by' rules (fixes #286)

### LINKED ISSUE #286: Deprecated/removed rules: Multiple replacements could use more padding or separator
In the [removed rules](http://eslint.org/docs/rules/#removed) section of the rules page, rules with multiple replacements show both replacements as `<a>` tags with no separator or padding in between.

For example, no-arrow-condition is replaced by no-confusing-arrow and no-constant-condition. But the replacement rules rule pretty closely together with only a space between them.

Possible solutions:

* Use an unordered list for rule replacements (most of the time it doesn't exceed two rules, so not a huge vertical space consumer)
* Use CSS to add padding between the links
* Use CSS or page generation to add content between each link (comma/space? forward slash? etc.)

vitorbal:
Add comma separators between the rules in the "replaced by" column of the deprecated/removed tables. Fixes #286.
Here's a preview:
screenshot 2016-09-14 10 54 54


styles/less/rules_index.less
```diff
@@ -10,3 +10,11 @@
         margin: 0;
     }
 }
+
+.replaced-by {
+    a + a:before {
+        content: ",";
+        display: inline-block;
```
vitorbal: @ilyavolodin small caveat: had to use `inline-block` instead of `inline`, otherwise the comma would get the hover effect of the link like this:
screenshot 2016-09-14 10 41 05

platinumazure:
LGTM (for what that's worth). Thanks for doing this!

ilyavolodin:
Ahh.. Sorry, totally forgot about this PR. Thanks for the reminder.

vitorbal:
Thanks, all! @ilyavolodin no worries :)

PR #296, opened 2016-10-04
Fix: Broken links to issue templates

daltones:
Extension `.md` was missing in a few URLs.


nzakas:
Thanks. We still need to fix our build system so this won't happen the next time we generate the site.

PR #297, opened 2016-10-08
Load jquery, bootstrap and anchor from CDNJS

PeterDaveHello:
Use CDNJS to load 3-party libraries with https + http/2.0 support, which is faster and safer.


ilyavolodin:
Sure,  why not. Thanks for doing this.

PR #307, opened 2016-10-28
ESLint 3.9.0 release notes

kaicataldo:


platinumazure:
Just noting that the codeframe examples look a little off (the caret is pointing about 2 characters to the right of where it should be).

PR #308, opened 2016-10-31
Docs: better default for deprecated rules with no replacement

vitorbal:
**What is the purpose of this pull request?**
Visually improve the way deprecated rules that have no replacement would look like in the "Deprecated" section of the rules list. With this change, the current list would look like this:

screenshot 2016-10-31 17 30 43

**What changes did you make? (Give an overview)**
Changed the rules list template to treat rules with an empty `replacedBy` property differently.

**Is there anything you'd like reviewers to focus on?**
We could also just leave the "replaced by" column empty instead, but I thought it looks like the table was a bit broken.. with this change, there is a small visual cue that the rule intentionally does not have any replacements.

### DIFF
--- a/docs/rules/index.md
+++ b/docs/rules/index.md
@@ -65,7 +65,11 @@ The `--fix` option on the [command line](../user-guide/command-line-interface#fi
 <tr>
 <td markdown="1">[{{rule.name}}]({{rule.name}})
 </td>
+{% if rule.replacedBy.size > 0 %}
 <td class="replaced-by" markdown="1">{% for replaced in rule.replacedBy %}[{{replaced}}]({{replaced}}){% endfor %}
+{% else %}
+<td class="replaced-by" markdown="1"><p class="text-muted">(no replacement)</p>
+{% endif %}
 </td>
 </tr>
 {% endfor %}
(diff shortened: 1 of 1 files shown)


nzakas:
I'm unsure if the dash makes it clear that there is no replacement. Maybe just "(no replacement)" would be better?

Don't forget to sign the new CLA. :)

ilyavolodin:
I actually prefer a dash. If you put any text in there, at a glance, users will think that every rule has a replacement. Dash clearly specifies that this field is left empty on purpose.

platinumazure:
We could certainly use "title" attribute or similar to get extra information in there if we need to. (Maybe we should think about accessibility here, as well...)

vitorbal:
A few more options, based on feedback:
screenshot 2016-11-03 14 26 07
screenshot 2016-11-03 14 28 28
screenshot 2016-11-03 14 30 00

ilyavolodin:
I still like the bottom one the most. But I could live with the second example.

nzakas:
Definitely not liking the title approach. I think having "no replacement" in parens and not linked is visually distinct enough not to cause confusion. Whether or not it's gray, I don't really care.

vitorbal:
Thanks for the feedback, everyone! I just pushed a new version with "(no replacement)" in grey, so everyone is happy :)
I also made sure the same default is applied to the "removed" section as well.

PR #317, opened 2016-12-15
Demo rewrite

ilyavolodin:
I think I got this into a reasonable shape to merge it into the site. It can definitely be improved a lot more. But I figured it would be easier to get smaller PRs from others, then sitting on the incomplete demo.

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -2,70 +2,8 @@
 title: ESLint Demo
 layout: demo
 ---
+
 <link rel="stylesheet" property="stylesheet" href="../styles/demo.css"/>
 
-<div class="container editorRow">
-    <div class="row">
-        <div class="col-md-7">
-            <pre id="editor" data-editor-lang="js" style="height:350px">var foo = bar;</pre>
-        </div>
-        <div class="col-md-5">
-            <div id="results">
-
-            </div>
-        </div>
-    </div>
-    <div class="row">
-        <div class="panel-group" id="accordion" role="tablist" aria-multiselectable="true">
-            <div class="panel panel-default">
-                <div class="panel-heading" role="tab" id="headingOne">
-                    <h4 class="panel-title">
-                        <a data-toggle="collapse" data-parent="#accordion" href="#configuration" aria-expanded="true" aria-controls="configuration">
-                            Rules Configuration
-                        </a>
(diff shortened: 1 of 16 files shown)


platinumazure:
This looks awesome!

Is there any way we can play with this (ideally without having to setup Jekyll locally)? Maybe is it worth putting the demo on a new experimental demo page?

ilyavolodin:
@platinumazure You can clone the repository and manually modify /demo/index.html page to insert header and footer from _includes directory. After that you should be able to serve it using any webserver (I use http-server for that).

not-an-aardvark:
I tried publishing a [GitHub pages build](https://not-an-aardvark.github.io/eslint.github.io/demo/) on my fork, but it didn't work very well because it's looking for styles from the root URL.

alberto:
@ilyavolodin  Good job! :+1: 

I think there is some room for improvement:
- it's not very idiomatic React in the use of event bus and excesive use of state
- there are a few minor issues here and there

Also it would be better for contributors if we used a more stardard stack in the React world (I know you are not a fan of them). Maybe jekyll can be a barrier here too (I know how to set it up and make it run incrementally, otherwise it's unbearable).

But, given it works fine and it's better than what we have now, we could merge it and gradually improve it later. Let me know if you agree or if you prefer to polish it first and want me to go over it.

ilyavolodin:
@alberto I don't really have anything against more standard React stack (I assume you are talking about ES6+, Redux and Webpack). But they require build process. And unless somebody can come up with a very simple build process that deploys to Github pages, I'm trying to avoid adding complexity in.

As I mentioned before, this is the first time I looked at React in any depth, so I expect there are a lot of things that can be improved here. So I think merging this in and polishing it incrementally is the best way to move this forward.

hzoo:
I think I mentioned this before but you can totally just use https://github.com/babel/babel-standalone which is basically a webpack version of babel in the browser for the demo page. We use this on the Babel REPL - https://babeljs.io/repl/.

Using unpkg - https://github.com/babel/babel.github.io/blob/6dcd843551548de71ff6f7568974dbdbf2375524/repl.html#L10

ilyavolodin:
Yes, that's my fallback strategy if people insist on using ES6:-)

PR #335, opened 2017-03-26
Update Espree Parser Configuration

Hirse:
espree-config


ilyavolodin:
Thanks for PR. index.build.js is a generated file, but I think it was manually generated by @xjamundx So I think it's fine to modify it by hands for now.

Hirse:
Yes, it is generated.
I generated it using the `npm run build` command.

ilyavolodin:
Ah, very nice. Thanks!

PR #340, opened 2017-04-06
Docs: Add "Getting Started" link to User Guid menu

platinumazure:
Given how important this page is for newcomers, we should ensure ease of access by including it in the User Guide dropdown. (Currently, it's linked to via a big button on the homepage, but the homepage doesn't have much other useful information and it's not hard to envision someone jumping straight to Configuring or the rules index.)


platinumazure:
@not-an-aardvark Dang. Unfortunately I drafted this via the GitHub interface, so I don't know if I can edit the commit message or not. I'll have to edit it as I merge.

PR #343, opened 2017-04-07
Fix blog links and rules page for 4.0

ilyavolodin:


docs/4.0.0/rules/index.md
```diff
@@ -0,0 +1,117 @@
+---
```
not-an-aardvark: Is this file just a copy of `docs/rules/index.md`? If so, could it be replaced with a symlink?

ilyavolodin: It's almost a copy. It uses different data source. This is so we would have different set of rules for 4.0 and 3.x

docs/4.0.0/rules/index.md
```diff
@@ -0,0 +1,117 @@
...
+layout: doc
+---
+
+# Rules
+
+Rules in ESLint are grouped by category to help you understand their purpose.
+
+No rules are enabled by default. The `"extends": "eslint:recommended"` property in a [configuration file](../user-guide/configuring#extending-configuration-files) enables rules that report common problems, which have a check mark (recommended) below.
+
+The `--fix` option on the [command line](../user-guide/command-line-interface#fix) automatically fixes problems (currently mostly whitespace) reported by rules which have a wrench (fixable) below.
```
not-an-aardvark: Will this be a 404 because the path is different?

ilyavolodin: Hard to figure out what you were pointing to, but I think it should be all fine, since it's relative paths. All of the documentation was copied into 4.0 folder. So `user-guide` folder is there too.

not-an-aardvark: Oh right, forgot about that. Ok, sounds good.

PR #344, opened 2017-04-07
Ensure anchors are positioned correctly

not-an-aardvark:
Currently, when an anchor targets an `a` tag, the page scrolls to the wrong location because the sticky header gets in the way. [Example](http://eslint.org/docs/4.0.0/user-guide/migrating-to-4.0.0#config-validation)


not-an-aardvark:
By the way, should we add `main.css.map` to gitignore? It seems like that file contains local path information, so it wouldn't work for anyone else without rebuilding anyway.

ilyavolodin:
Sure, sounds good to me.

PR #361, opened 2017-06-24
Save demo state in localStorage

not-an-aardvark:
This restores the save-state behavior of the old editor rather than reverting to the default state every time.

### DIFF
--- a/js/app/demo/app.jsx
+++ b/js/app/demo/app.jsx
@@ -15,7 +15,7 @@ define(['react', 'jsx!editor', 'jsx!messages', 'jsx!fixedCode', 'jsx!configurati
                 return result;
             }
 
-            return {
+            var initialState = JSON.parse(window.localStorage.getItem('linterDemoState')) || {
                 messages: [],
                 options: {
                     parserOptions: {
@@ -68,6 +68,9 @@ define(['react', 'jsx!editor', 'jsx!messages', 'jsx!fixedCode', 'jsx!configurati
                 fix: false,
                 fixedText: 'var foo = bar;'
             };
+
+            this.initialText = initialState.text;
+            return initialState;
         },
 
         handleChange: function(event) {
@@ -110,12 +113,15 @@ define(['react', 'jsx!editor', 'jsx!messages', 'jsx!fixedCode', 'jsx!configurati
                 });
             }.bind(this))
         },
+        componentDidUpdate: function() {
+            window.localStorage.setItem('linterDemoState', JSON.stringify(this.state));
+        },
         render: function() {
             return (
                 <div className="container editorRow">
                     <div className="row">
                         <div className="col-md-7">
-                            <Editor onChange={this.handleChange} text="var foo = bar;" errors={this.state.messages} />
+                            <Editor onChange={this.handleChange} text={this.initialText} errors={this.state.messages} />
                         </div>
                         <div className="col-md-5">
                             <ul className="nav nav-tabs" role="tablist">
--- a/js/app/demo/app.jsx
+++ b/js/app/demo/app.jsx
@@ -1,6 +1,24 @@
 'use strict';
 
-define(['react', 'lodash', 'jsx!editor', 'jsx!messages', 'jsx!fixedCode', 'jsx!configuration', 'eslint'], function(React, lodash, Editor, Messages, FixedCode, Configuration, Linter) {
+define(['react', 'jsx!editor', 'jsx!messages', 'jsx!fixedCode', 'jsx!configuration', 'eslint'], function(React, Editor, Messages, FixedCode, Configuration, Linter) {
+    function defaultsDeep(object, defaults) {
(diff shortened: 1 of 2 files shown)


js/app/demo/app.jsx
```diff
@@ -110,12 +113,15 @@ define(['react', 'jsx!editor', 'jsx!messages', 'jsx!fixedCode', 'jsx!configurati
...
+        componentDidUpdate: function() {
+            window.localStorage.setItem('linterDemoState', JSON.stringify(this.state));
+        },
         render: function() {
             return (
                 <div className="container editorRow">
                     <div className="row">
                         <div className="col-md-7">
-                            <Editor onChange={this.handleChange} text="var foo = bar;" errors={this.state.messages} />
+                            <Editor onChange={this.handleChange} text={this.initialText} errors={this.state.messages} />
```
not-an-aardvark: I'm not sure exactly what's going on, but when I set this to `text={this.state.text}`, the editor breaks and stops being editable after entering a single character. However, it is editable if the prop is static.

ilyavolodin: Most likely if you add it to dynamic state, whenever you type a letter, React tries to reset it back to old state, then changes state to what you typed, then tries to reset it to new state, etc.

js/app/demo/app.jsx
```diff
@@ -15,7 +15,7 @@ define(['react', 'jsx!editor', 'jsx!messages', 'jsx!fixedCode', 'jsx!configurati
                 return result;
             }
 
-            return {
+            var initialState = JSON.parse(window.localStorage.getItem('linterDemoState')) || {
```
ilyavolodin: This will result in the same issues that we had with the old behavior (i.e. every time we add a new rule, it doesn't show up in the demo, until you manually wipe out localStorage). We should either add version number, and invalidate localStorage on version change, or merge state from localStorage with new rules (might still result in bugs).

not-an-aardvark: Good point. I think merging state would be better -- it seems annoying to invalidate it every time a new version comes out. (Personally I would want to set a configuration once and then never have to deal with the default configuration again.)

js/app/demo/app.jsx
not-an-aardvark: That makes sense. Do you know if there's a way to avoid this? (I know this works fine with `<textarea>` elements.) It doesn't have to be resolved right now, but it seems like storing the state of the editor in two different components goes against the React principle of having a single source of truth.

smably: Passing a static value as the `text` prop only works by virtue of React's virtual DOM diffing: even though changes to the text in the editor cause the parent component to update its state and re-render, the text isn't reset because React identifies that the rendered output from `Editor` is identical and avoids recreating the DOM node, thus preserving the text.

One simple solution is to add a `shouldComponentUpdate` check to the `Editor` component and always return false, since the `text` prop is only used to seed the initial value and any subsequent updates happen outside React's render lifecycle.

not-an-aardvark: That makes sense. I think my confusion was that I originally expected the `text` prop to always be an up-to-date representation of the text, when it is actually just the initial text. Maybe that prop should be renamed.

ilyavolodin: I think in order for it to work as you were expecting, Editor should become a controlled component. But in all honesty, I'm not sure. Demo is the only time I used React, so....:-)

js/app/demo/app.jsx
ilyavolodin: Merging states would probably be a bit better, but will cause issues when options for a given rule change.

not-an-aardvark: We don't support options in the demo, so I don't think this is a problem.

not-an-aardvark:
Updated to merge options rather than replacing.

I added lodash as a script dependency for this because it seems like writing our own recursive merge function probably isn't worthwhile. I can remove it if anyone has objections to doing this.

ilyavolodin:
I think we need to extract just `defaultsDeep` function from lodash. All of the dependencies have to be downloaded by the client on every demo load. And lodash is 25k minified and gziped, which is not a small amount at all. Also, could you add checks around localStorage access? I think localStorage is not available for privacy browsing.

not-an-aardvark:
> I think localStorage is not available for privacy browsing.

I'm pretty sure it's still available, it just uses a separate storage instance.

smably:
I believe Safari in private browsing mode will throw an error if you try to write to localStorage: https://stackoverflow.com/questions/14555347/html5-localstorage-error-with-safari-quota-exceeded-err-dom-exception-22-an

This is the method I've used to check whether it's available: https://gist.github.com/paulirish/5558557#stronger-full-capability-test-for-localstorage-with-ios-private-browsing-protection

not-an-aardvark:
Huh, that's odd. Ok, I'll add a check for it.

PR #367, opened 2017-07-10
Add a 404 page and an “Edit this page” link, as well as some other polish

j-f1:
[Live demo](https://j-f1.github.io/eslint.github.io/)

Here’s the 404 page on desktop and mobile: (click to enlarge)

desktopmobile

In this PR, I’ve:
* added a CSS sourcemap
* added an “Edit this page” link to the footers of each page, linking to the appropriate page to edit the file directly from GitHub
* added a popup banner that asks you to update to the latest version of ESLint when you’re browsing outdated docs
* aligned the ESLint logo vertically with the “ESLint” text: (before is on the left, after is on the right)
   alterations
* Used `site.url` in HTML links

If you’d like me to change anything, please ask and I’ll try my best 💯

Depends on eslint/eslint#9049 for proper operation.


j-f1:
It looks like the Netlify deploy won’t work because there’s no way to change `site.url` in the config dynamically.

styles/main.css.map
```diff
@@ -0,0 +1 @@
...AA,CACA,iBAzFF,UA2FC,KACC,aAAA,CACA,iBA7FF,UA+FC,IACC,cAhGF,UAkGC,IACC,cAnGF,UAqGC,KACC,cAtGF,UAwGC,KACC,cAzGF,UA2GC,KACC,aAAA,CACA,iBA7GF,UA+GC,KACC,cAhHF,UAkHC,KACC,aAAA,CACA,iBApHF,UAsHC,KACC,cAvHF,UAyHC,KACC,aAAA,CACA,iBA3HF,UA6HC,KACC,aAAA,CACA,iBA/HF,UAiIC,KACC,aAAA,CACA,iBAnIF,UAqIC,KACC,cAtIF,UAwIC,KACC,cAzIF,UA2IC,KACC,cA5IF,UA8IC,KACC,aAAA,CACA,iBAhJF,UAkJC,IACC,cAnJF,UAqJC,KACC,cAtJF,UAwJC,KACC,cAzJF,UA2JC,KACC,cA5JF,UA8JC,KACC,cA/JF,UAiKC,KACC,cAlKF,UAoKC,KACC,cArKF,UAuKC,KACC,cAxKF,UA0KC,KACC,cA3KF,UA6KC,KACC,cA9KF,UAgLC,KACC,cAjLF,UAmLC,KACC,cApLF,UAsLC,KACC,cAvLF,UAyLC,KACC,cA1LF,UA4LC,KACC,cA7LF,UA+LC,KACC,cAhMF,UAkMC,KACC,cAnMF,UAqMC,KACC,cAtMF,UAwMC,KACC,cAzMF,UA2MC,KACC,cA5MF,UA8MC,KACC,cCpOF,UACI,cADJ,UACkB,UACV,SAFR,UAII,OACI,UALR,UAQI,GACI,SAIR,YACI,EAAE,EAAG,QACD,QAAS,GAAT,CACA,oBAAA,CACA,iBCjBR,qBACE,cAAA,CACA,mBAEA,QAA+C,0BA+BjD,qBA9BI,cA8BJ,oBA5BI,IACE,gBA2BN,oBAzBI,IACE,kBAIJ,qBAfF,oBAeK,IAfL,oBAeW,IAfX,oBAeiB,IAfjB,oBAeuB,IAfvB,oBAe6B,IACzB,kBADF,oBAGE,gBAlBJ,oBAeK,GAGD,gBAlBJ,oBAeW,GAGP,gBAlBJ,oBAeiB,GAGb,gBAlBJ,oBAeuB,GAGnB,gBAlBJ,oBAe6B,GAGzB,gBACE,aAnBN,oBAuBE,EAAI,sBACF,cAAA,CACA,gBAAA,CACA,kBA1BJ,oBA6BE,KACE,WAAA,CACA,uBAAA,CACA","file":"main.css"}
```
not-an-aardvark: Can you remove this sourcemap? I think we had a sourcemap in the repo at one point, but we removed it because it only contains user-specific info. For example, this file contains paths on your local filesystem, so it wouldn't work for other people.

_includes/footer.html
```diff
@@ -5,6 +5,11 @@
 
       <footer>
         <ul class="list-inline">
+          {% if page.path contains 'docs' %}
```
not-an-aardvark: I'm not sure this check is reliable. For example, the edit button at https://j-f1.github.io/eslint.github.io/docs/developer-guide/contributing/ goes to a 404 because it links to `index.md` rather than `README.md`.

It also feels a bit fragile to do a substring check -- for example, if we added a page to `eslint.github.io` called `how-to-write-docs.md`, it would end up linking to the `eslint` repo.

_layouts/doc.html
```diff
@@ -2,6 +2,13 @@
 {% include menu.html %}
 <main class="doc">
   <article class="container">
+    <div class="alert alert-warning hidden alert-outdated" role="alert">
+      <strong class="header">Heads up!</strong>
+      <p>
+        This document is for an old version of ESLint that is <em>no longer supported</em>.
+        Please <a class="alert-link" href="{{ site.url }}/docs/user-guide/migrating-to-4.0.0">upgrade</a> to a newer release!
```
not-an-aardvark: I don't mind doing this, but I think people typically just use the current version of the docs anyway, and only use the versioned URL when they deliberately need to look at the docs for an older version. So the message might not be useful to many people since they would already know they're using an unsupported version.

_includes/footer.html
j-f1: Unfortunately, Liquid doesn’t to RegExps. Should I change it to `/docs`, or should I generate the link client-side?

styles/main.css.map
j-f1: Gone 🔥

_layouts/doc.html
j-f1: Removed 🔥

_includes/menu.html
```diff
@@ -2,7 +2,7 @@
 <header class="navbar navbar-default navbar-demo navbar-fixed-top eslint-nav" id="top" role="banner">
   <div class="container">
 
-    <a href="/" class="navbar-brand"><img alt="ESLint" src="/img/logo.svg" itemprop="image">ESLint</a>
+    <a href="{{ site.url }}/" class="navbar-brand"><img alt="ESLint" src="{{ site.url }}/img/logo.svg" itemprop="image">ESLint</a>
```
ilyavolodin: What's the reason for changing everything to use absolute path? While they will be gzipped, they will still add to download size of the page, and I'm not sure I know the advantages of using them in this case.

_includes/footer.html
ilyavolodin: Nice change. But I think this is better done through our build script in the main repo. We already add things like version added and such in that script. Adding a link to edit the page should be pretty easy there, and wouldn't require string comparison.

_includes/menu.html
j-f1: When editing a forked version, the site isn’t hosted at the root of the domain, so adding these allows the site to load the assets properly.

ilyavolodin: But all of those path are starting with `/` so if you forked the whole repo, you probably have all of the assets in that repo as well, right? I guess all of us always just used local install of Jekyll to work on the site, and never tried it in a fork.

not-an-aardvark: No, because the site would end up at `https://<username>.github.io/eslint.github.io`, so the path `/img/logo.svg` would load the image from `https://<username>.github.io/img/logo.svg`, rather than `https://<username>.github.io/eslint.github.io/img/logo.svg`.

I've also encountered this issue when trying to demo things for this repo.

ilyavolodin: Ah, got it. Thanks. As I mentioned, I always used local Jekyll, and that's not a problem there, since it's always `http://localhost:port/` in that case, I'm 👍 for adding that in.

ilyavolodin:
We don't really need to worry about Netlify. That was just an experiment, we don't use it for anything.

j-f1:
Should I open a PR for that on eslint/eslint?

ilyavolodin:
> Should I open a PR for that on eslint/eslint?

I think so. I think it will be much cleaner change there.

j-f1:
@ilyavolodin @not-an-aardvark Updated. Sorry for the delay.

docs/about/index.md
```diff
@@ -1,6 +1,7 @@
 ---
 title: About
 layout: doc
+edit_link: https://github.com/eslint/eslint/edit/master/lib/rules/index.js
```
not-an-aardvark: This doesn't seem correct -- the link goes to a 404.

j-f1:
@not-an-aardvark Done!

not-an-aardvark:
Thanks for contributing!

PR #375, opened 2017-08-18
Update: url to https in config

gyandeeps:


gyandeeps:
I am guessing most of the places use `{site.urll}`... if not then we can do it case by case.

PR #381, opened 2017-08-23
Avoid unnecessary state in demo

not-an-aardvark:
The list of problems and output can be considered a pure function of the text and options, so it's not necessary to store it separately as state and update it whenever the text or options change. Instead, just calculate it in the `render()` method.


gyandeeps:
so are we always fixing?

btmills:
> so are we always fixing?

@gyandeeps it pulls the `fix` option from `state` here:

https://github.com/eslint/eslint.github.io/blob/d27806812c02ac50823f4ea23a0db20301778808/js/app/demo/app.jsx#L111

PR #382, opened 2017-08-23
Store demo tab state in react instead of bootstrap

not-an-aardvark:
This fixes an issue where the state from bootstrap is sometimes out of sync with the state stored in `localStorage`, resulting in the "messages" tab being active while displaying the fixed code instead.

### DIFF
--- a/js/app/demo/app.jsx
+++ b/js/app/demo/app.jsx
@@ -110,11 +110,6 @@ define(['react', 'jsx!editor', 'jsx!messages', 'jsx!fixedCode', 'jsx!configurati
         lint: function() {
             return linter.verifyAndFix(this.state.text, this.state.options, { fix: this.state.fix });
         },
-        componentDidMount: function() {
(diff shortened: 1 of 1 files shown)


js/app/demo/app.jsx
```diff
@@ -157,8 +150,41 @@ define(['react', 'jsx!editor', 'jsx!messages', 'jsx!fixedCode', 'jsx!configurati
...
+                                    className={this.state.fix ? "" : "active"}
+                                >
+                                    <a
+                                        href="#messages"
+                                        aria-controls="messages"
+                                        role="tab"
+                                        onClick={
+                                            function() {
+                                                this.setState({ fix: false });
+                                            }.bind(this)
```
btmills: The value of these `onClick` props will be reference-unequal every time this component renders since it's binding a new function very time. What would you think of converting these to handlers on the class? The real performance impact will be negligible, but I wanted to point it out anyway.

PR #385, opened 2017-08-23
Avoid duplicate state in parserOptions

not-an-aardvark:
The ParserOptions component contains duplicated state for the active ESLint configuration, but the state should be only stored into the main app and passed down as props, rather than being stored in ParserOptions. This updates ParserOptions to be a functional component instead.

### DIFF
--- a/js/app/demo/parserOptions.jsx
+++ b/js/app/demo/parserOptions.jsx
@@ -1,59 +1,72 @@
 'use strict';
 
 define(['react'], function(React) {
-    return React.createClass({
-        displayName: 'ParserOptions',
-        getInitialState: function() {
-            return this.props.options;
-        },
-        handleEcmaVersion: function(event) {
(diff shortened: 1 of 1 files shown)


js/app/demo/parserOptions.jsx
```diff
@@ -1,59 +1,72 @@
...
+                                            <input
+                                                type="checkbox"
+                                                checked={props.options.ecmaFeatures[ecmaFeature]}
+                                                className="option-checkbox"
+                                                id={ecmaFeature}
+                                                onChange={
+                                                    function(event) {
+                                                        var updatedFeature = {};
+                                                        updatedFeature[ecmaFeature] = event.target.checked;
+                                                        props.onChange(Object.assign({}, props.options, {
```
btmills: This should be `onUpdate`

Issue #99, opened 2015-07-16
Demo: an ability to specify code with URL.


mysticatea:
I want to use demo to report issues.
I'm happy if I can specify code with URL.

Such as:

```
http://eslint.org/demo/#code=var%20foo%20%3D%20%22Hello%20World!%22
```

``` js
var foo = "Hello World!";
```

gyandeeps:
I like it. Its so handy when I want to share it with people.
I know esprima does it on their website.

ilyavolodin:
Yes, it's on my todo list. It's just not that simple, because rule configuration matters a lot. And we have a lot of rules and options. I don't think it all will fit on query string, so we probably need to get a database to store configuration + code and output a token.

nzakas:
Eek, I really wanted to keep the site static for simplicity - otherwise we will be spending time debugging why the DB connection is slow, broken, etc.

ilyavolodin:
I wasn't planning on making the site dynamic. Just exposing a web service,
that can be called to save and retrieve examples.

nzakas:
Yup, that's what I was referring to...why introduce something that could break? The demo isn't really meant to be anything more than just that, a demo to try it out. I think doing anything more than the query string solution is a lot of work for very little gain and maintenance headaches.

ilyavolodin:
My idea for this change was to turn demo into a bug reporting tool as well. It would be much easier and quicker to ask people to reproduce their issue in the demo, and past URL into github issue, then chase them for version of ESLint, .eslintrc and error messages that they are seeing.
I originally hoped to find some free service that just allows arbitrary key-value pair storage and retrieval, but I couldn't find anything out there that does this.

guerrero:
I think that it can be easy to implement this functionality with JavaScript, due that Jekyll is a static site generator there's no other way to achieve it. We can use a library like [URI.js](https://medialize.github.io/URI.js/) or use plain vanilla JavaScript and `window.location.href`. But the steps are basically the same:

To share a link:
- Iterate over the rules to check which of them are active and store them in a dictionary
- Get the content of the editor and store it in another variable
- Convert them to an URL query using a library
- Optionally we can add an eventListener to update the website when the editor content or a rule changes but I think that it wouldn't be a good idea

To get info from a URL query:
- Parse the query and to the inverse process but checking before if a rule exists and if that's not the case append all the errors to a notification panel that can be placed above the editor

[...]

PR #389, opened 2017-08-25
Allow a permalink to the demo with a given piece of code and config

not-an-aardvark:
Fixes https://github.com/eslint/eslint.github.io/issues/99

This updates the demo to allow code and config to be specified with a URL. Hopefully, this should make the demo an easy bug-reporting tool -- as an alternative to filling out information on Node versions etc., users could just paste a link to the demo with a particular code and config to demonstrate the bug.

The serialization format is:

`atob(JSON.stringify({ text: "the text", options: { /* the config */ })`

The serialized state is placed in the URL hash whenever the config or text is changed.

This commit also updates the config-building logic to make the serialized configs smaller, to prevent the URLs from getting too long. In particular, rules that were previously configured with `"off"` are now omitted from the config, and envs that were previously `false` are now omitted.

Even with these changes, the URL for the default state is still very long, because every rule from `eslint:recommended` is explicitly enabled in the demo config by default:

<details>
<summary>URL for default text/config</summary>

```
https://eslint.org/demo/#eyJ0ZXh0IjoidmFyIGZvbyA9IGJhcjsiLCJvcHRpb25zIjp7InBhcnNlck9wdGlvbnMiOnsiZWNtYVZlcnNpb24iOjUsInNvdXJjZVR5cGUiOiJzY3JpcHQiLCJlY21hRmVhdHVyZXMiOnsianN4IjpmYWxzZSwiZ2xvYmFsUmV0dXJuIjpmYWxzZSwiaW1wbGllZFN0cmljdCI6ZmFsc2UsImV4cGVyaW1lbnRhbE9iamVjdFJlc3RTcHJlYWQiOmZhbHNlfX0sInJ1bGVzIjp7ImNvbnN0cnVjdG9yLXN1cGVyIjoiZXJyb3IiLCJuby1jYXNlLWRlY2xhcmF0aW9ucyI6ImVycm9yIiwibm8tY2xhc3MtYXNzaWduIjoiZXJyb3IiLCJuby1jb21wYXJlLW5lZy16ZXJvIjoiZXJyb3IiLCJuby1jb25kLWFzc2lnbiI6ImVycm9yIiwibm8tY29uc29sZSI6ImVycm9yIiwibm8tY29uc3QtYXNzaWduIjoiZXJyb3IiLCJuby1jb25zdGFudC1jb25kaXRpb24iOiJlcnJvciIsIm5vLWNvbnRyb2wtcmVnZXgiOiJlcnJvciIsIm5vLWRlYnVnZ2VyIjoiZXJyb3IiLCJuby1kZWxldGUtdmFyIjoiZXJyb3IiLCJuby1kdXBlLWFyZ3MiOiJlcnJvciIsIm5vLWR1cGUtY2xhc3MtbWVtYmVycyI6ImVycm9yIiwibm8tZHVwZS1rZXlzIjoiZXJyb3IiLCJuby1kdXBsaWNhdGUtY2FzZSI6ImVycm9yIiwibm8tZW1wdHktY2hhcmFjdGVyLWNsYXNzIjoiZXJyb3IiLCJuby1lbXB0eS1wYXR0ZXJuIjoiZXJyb3IiLCJuby1lbXB0eSI6ImVycm9yIiwibm8tZXgtYXNzaWduIjoiZXJyb3IiLCJuby1leHRyYS1ib29sZWFuLWNhc3QiOiJlcnJvciIsIm5vLWV4dHJhLXNlbWkiOiJlcnJvciIsIm5vLWZhbGx0aHJvdWdoIjoiZXJyb3IiLCJuby1mdW5jLWFzc2lnbiI6ImVycm9yIiwibm8tZ2xvYmFsLWFzc2lnbiI6ImVycm9yIiwibm8taW5uZXItZGVjbGFyYXRpb25zIjoiZXJyb3IiLCJuby1pbnZhbGlkLXJlZ2V4cCI6ImVycm9yIiwibm8taXJyZWd1bGFyLXdoaXRlc3BhY2UiOiJlcnJvciIsIm5vLW1peGVkLXNwYWNlcy1hbmQtdGFicyI6ImVycm9yIiwibm8tbmV3LXN5bWJvbCI6ImVycm9yIiwibm8tb2JqLWNhbGxzIjoiZXJyb3IiLCJuby1vY3RhbCI6ImVycm9yIiwibm8tcmVkZWNsYXJlIjoiZXJyb3IiLCJuby1yZWdleC1zcGFjZXMiOiJlcnJvciIsIm5vLXNlbGYtYXNzaWduIjoiZXJyb3IiLCJuby1zcGFyc2UtYXJyYXlzIjoiZXJyb3IiLCJuby10aGlzLWJlZm9yZS1zdXBlciI6ImVycm9yIiwibm8tdW5kZWYiOiJlcnJvciIsIm5vLXVuZXhwZWN0ZWQtbXVsdGlsaW5lIjoiZXJyb3IiLCJuby11bnJlYWNoYWJsZSI6ImVycm9yIiwibm8tdW5zYWZlLWZpbmFsbHkiOiJlcnJvciIsIm5vLXVuc2FmZS1uZWdhdGlvbiI6ImVycm9yIiwibm8tdW51c2VkLWxhYmVscyI6ImVycm9yIiwibm8tdW51c2VkLXZhcnMiOiJlcnJvciIsIm5vLXVzZWxlc3MtZXNjYXBlIjoiZXJyb3IiLCJyZXF1aXJlLXlpZWxkIjoiZXJyb3IiLCJ1c2UtaXNuYW4iOiJlcnJvciIsInZhbGlkLXR5cGVvZiI6ImVycm9yIn0sImVudiI6e319fQ==
```
</details>

It might be possible to make this shorter in the default case by making the config extend from `eslint:recommended`, and then explicitly disabling rules from `eslint:recommended` in the config if necessary.

Security considerations:

* If it's possible to cause ESLint to execute custom code when linting a given text with a given config, then this change will allow [reflected XSS](https://www.owasp.org/index.php/Cross-site_Scripting_(XSS)) on the `eslint.org` origin. An attacker could get a victim to click a link/load an iframe with a URL that puts malicious code in the demo, and some JavaScript would get run in the user's browser. `eslint.org` is a static site, so it would be difficult for an attacker to cause any significant damage, but reflected XSS is still undesirable. This change assumes that it's not possible for this to happen when linting untrusted code.
* An attacker could tamper with the user's cached state in `localStorage` by getting the user to click a link/load an iframe with different code/config. This could be mildly annoying for the user. To avoid this, if code is loaded from the URL hash, the state is not saved to `localStorage` until the user modifies the code or the config.

### DIFF
--- a/js/app/demo/app.jsx
+++ b/js/app/demo/app.jsx
@@ -86,12 +86,7 @@ define(['react', 'jsx!editor', 'jsx!messages', 'jsx!fixedCode', 'jsx!configurati
                     parserOptions: {
                         ecmaVersion: 5,
                         sourceType: 'script',
-                        ecmaFeatures: {
-                            jsx: false,
-                            globalReturn: false,
-                            impliedStrict: false,
-                            experimentalObjectRestSpread: false
-                        }
+                        ecmaFeatures: {}
                     },
                     rules: (function() {
                         var result = {};
--- a/js/app/demo/app.jsx
+++ b/js/app/demo/app.jsx
@@ -31,6 +31,7 @@ define(['react', 'jsx!editor', 'jsx!messages', 'jsx!fixedCode', 'jsx!configurati
 
     var linter = new Linter();
     var rules = linter.getRules();
+    var ruleNames = Array.from(rules.keys());
     var docs = (function() {
         var map = rules;
         var result = {};
@@ -39,12 +40,48 @@ define(['react', 'jsx!editor', 'jsx!messages', 'jsx!fixedCode', 'jsx!configurati
         });
         return result;
     })();
+
+    var ENV_NAMES = [
+        "browser",
+        "node",
+        "commonjs",
+        "shared-node-browser",
+        "worker",
+        "amd",
+        "mocha",
+        "jasmine",
+        "jest",
+        "phantomjs",
+        "jquery",
+        "qunit",
+        "prototypejs",
+        "shelljs",
+        "meteor",
+        "mongo",
+        "protractor",
+        "applescript",
+        "nashorn",
+        "serviceworker",
+        "atomtest",
+        "embertest",
+        "webextensions",
+        "es6",
+        "greasemonkey"
+    ];
+
     return React.createClass({
         displayName: 'App',
         getInitialState: function() {
             var storedState = JSON.parse(window.localStorage.getItem('linterDemoState') || '{}');
+            var urlState = (function() {
+                try {
+                    return JSON.parse(window.atob(window.location.hash.replace(/^#/, "")));
+                } catch (err) {
+                    return null;
+                }
+            }());
 
-            var initialState = defaultsDeep(storedState, {
+            var initialState = defaultsDeep(urlState || storedState, {
                 options: {
                     parserOptions: {
                         ecmaVersion: 5,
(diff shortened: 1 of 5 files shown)


ilyavolodin:
We could do something like what astexplorer.net (or was that some other online tool?) is doing and store the code/config in gist.

not-an-aardvark:
That's true. One potential issue with using gists is that unlike astexplorer.net, we don't have any backend to store OAuth tokens, and I'm unsure about whether the GitHub API supports anonymous OAuth tokens (i.e. creating a gist without authenticating through the user's account first).

PR #393, opened 2017-08-26
Lint JS files

not-an-aardvark:
This adds a task to lint the JS files with `eslint-config-eslint` and some rules from `eslint-plugin-react` and `eslint-plugin-jsx-a11y`. It also fixes some existing linting errors.

The task can be run using `npm run lint` or `npm test`. It does not run on Travis yet. I wasn't sure how to make it run on Travis, because we already have the language set to ruby in `.travis.yml` to run jekyll, and I'm not sure of the best way to also have a JS build. Ideas on how to do that would be appreciated, although I would also be fine with merging this for now and figuring out how to make it run on Travis later.

I added disable comments liberally in existing code where it wasn't obvious how to fix the problem. The goal of this change is mostly to prevent future errors, although we can refactor old code later on.


js/app/demo/rulesConfig.jsx
```diff
@@ -26,12 +27,24 @@ define(['react'], function(React) {
...
+            if (e.target.checked) {
+                updatedConfig[key] = "error";
+            } else {
+                delete updatedConfig[key];
+            }
+            props.onUpdate(updatedConfig);
+        }
         function getRow(i) {
             var limit = Math.ceil(props.ruleNames.length / 3);
-            const start = limit * i;
```
not-an-aardvark: As a result of this error, the demo was unusable on non-ES6 browsers for a few days. :confused:

PR #396, opened 2017-08-26
Show end locations of problems in demo

not-an-aardvark:
For the code sample [here](https://eslint.org/demo/#eyJ0ZXh0IjoiLyogZXNsaW50IHJlcXVpcmUtanNkb2M6IGVycm9yLCBuby11bmRlZjogZXJyb3IsIGtleXdvcmQtc3BhY2luZzogZXJyb3IgKi9cblxuZm9vYmFyXG5cbnZhciBmb28gPSAxO1xuZm9vKCk7XG5cbmZ1bmN0aW9uIGJhcigpIHtcbiAgICByZXR1cm47XG59XG5cbmJhcigpO1xuIiwib3B0aW9ucyI6eyJwYXJzZXJPcHRpb25zIjp7ImVjbWFWZXJzaW9uIjo2LCJzb3VyY2VUeXBlIjoic2NyaXB0IiwiZWNtYUZlYXR1cmVzIjp7fX0sInJ1bGVzIjp7fSwiZW52Ijp7fX19):

Before

After

### DIFF
--- a/js/app/demo/app.jsx
+++ b/js/app/demo/app.jsx
@@ -118,12 +118,18 @@ define(["react", "jsx!editor", "jsx!messages", "jsx!fixedCode", "jsx!configurati
         },
         render: function App() {
             var results = this.lint();
+            var sourceCode = linter.getSourceCode();
 
             return (
                 <div className="container editorRow">
                     <div className="row">
                         <div className="col-md-7">
-                            <Editor onChange={this.handleChange} text={this.initialText} errors={results.messages} />
+                            <Editor
+                                onChange={this.handleChange}
+                                text={this.initialText}
+                                errors={results.messages}
+                                getIndexFromLoc={sourceCode && sourceCode.getIndexFromLoc.bind(sourceCode)}
+                            />
                         </div>
                         <div className="col-md-5">
                             <ul className="nav nav-tabs" role="tablist">
--- a/js/app/demo/editor.jsx
+++ b/js/app/demo/editor.jsx
@@ -45,14 +45,28 @@ define(["react", "orion", "reactDom", "events"], function(React, orion, ReactDOM
         showProblems: function() {
             if (this.props.errors) {
                 this.editor.showProblems(this.props.errors.map(function(error) {
+                    if (error.fatal) {
+                        return {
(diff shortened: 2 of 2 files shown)


js/app/demo/app.jsx
```diff
@@ -134,11 +134,17 @@ define(['react', 'jsx!editor', 'jsx!messages', 'jsx!fixedCode', 'jsx!configurati
         },
         render: function() {
             var results = this.lint();
+            var sourceCode = linter.getSourceCode();
```
not-an-aardvark: Using `getSourceCode` here is not an ideal solution, since it's not part of the public API the result changes after different `linter.verify` calls. It might be a good idea to expose a static method on `Linter` to convert range indices to locations, or to get the start/end indices of a reported problem.

js/app/demo/editor.jsx
```diff
@@ -50,10 +50,24 @@ define(["react", "orion", "reactDom", "events"], function(React, orion, ReactDOM
...
+                        line: error.endLine || error.line,
+                        column: (error.endColumn || error.column) - 1
+                    });
+
                     return {
-                        line: error.line,
-                        start: error.column,
-                        end: error.column + 1,
+                        start: start,
+                        end: start === end ? end + 1 : end,
```
btmills: Is there documentation for the shape of the objects passed to `showProblems` somewhere? The [docs I found](https://orionhub.org/jsdoc/symbols/orion.editor.Editor.html#showProblems) just say `showProblems(problems)`, which isn't very helpful. I'm only asking because the fatal error condition includes a `line` property and this doesn't, and I just wanted to make sure that's intentional.

not-an-aardvark: I couldn't find any, unfortunately -- I discovered this API by looking at [the source code](https://github.com/eclipse/orion.client/blob/ee564a2964ca2ead6cfe6752c13d50abf8b86c50/bundles/org.eclipse.orion.client.editor/web/orion/editor/editor.js#L1241-L1262). It appears that when `line` is provided, `start` and `end` are considered to be column offsets, otherwise `start` and `end` are considered to be global indices in the text.

PR #399, opened 2017-08-28
Fix incorrect falsy check for localStorage stored state

not-an-aardvark:
Previously, the function to get state from `localStorage` would return an empty object as the default value if there was nothing stored. However, another part of the code was checking whether the value was falsy. This was causing crashes when going to https://eslint.org/demo without any stored state or additional data in the URL.

This commit updates the `localStorage` handler to return `null` as the default value.


not-an-aardvark:
I'm going to merge this quickly because the bug is basically breaking the demo.

PR #400, opened 2017-08-28
Fix issue where "fix" state was sometimes unset

not-an-aardvark:
When loading code from a URL, the the value of `this.state.fix` would end up undefined. This would result in ESLint running in autofix mode, but the "fix" tab wouldn't get displayed, resulting in some missing problems.

[Example](https://eslint.org/demo/#eyJ0ZXh0IjoiZm9vIiwib3B0aW9ucyI6eyJwYXJzZXJPcHRpb25zIjp7ImVjbWFWZXJzaW9uIjo4LCJzb3VyY2VUeXBlIjoic2NyaXB0IiwiZWNtYUZlYXR1cmVzIjp7fX0sInJ1bGVzIjp7InNlbWkiOjJ9LCJlbnYiOnt9fX0=)

Note that there are initially no errors, but switching to the "fix" tab and back causes a `semi` error to appear.

### DIFF
--- a/js/app/demo/app.jsx
+++ b/js/app/demo/app.jsx
@@ -63,27 +63,29 @@ define(['react', 'jsx!editor', 'jsx!messages', 'jsx!fixedCode', 'jsx!configurati
                 }
             }());
 
-            var initialState = urlState || storedState || {
-                options: {
-                    parserOptions: {
-                        ecmaVersion: 5,
-                        sourceType: 'script',
-                        ecmaFeatures: {}
+            var initialState = Object.assign(
+                { fix: false },
+                urlState || storedState || {
+                    options: {
+                        parserOptions: {
+                            ecmaVersion: 5,
+                            sourceType: 'script',
+                            ecmaFeatures: {}
+                        },
+                        rules: (function() {
+                            var result = {};
+                            rules.forEach(function(rule, ruleId) {
+                                if (rule.meta.docs.recommended) {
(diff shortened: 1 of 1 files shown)


js/app/demo/app.jsx
```diff
@@ -63,27 +63,29 @@ define(['react', 'jsx!editor', 'jsx!messages', 'jsx!fixedCode', 'jsx!configurati
...
+                urlState || storedState || {
+                    options: {
+                        parserOptions: {
+                            ecmaVersion: 5,
+                            sourceType: 'script',
+                            ecmaFeatures: {}
+                        },
+                        rules: (function() {
+                            var result = {};
+                            rules.forEach(function(rule, ruleId) {
```
gyandeeps: nit: You can use reduce and not use variables at all.

not-an-aardvark: I agree, although this code was already here before -- this PR just modifies the indentation. I can make the fix to use reduce in a separate PR.

not-an-aardvark: Actually, it's nontrivial to do this because `rules` is a `Map`, not an array. So it would require something like

```js
rules: Array.from(rules).filter(function(ruleData) {
    return ruleData[1].meta.docs.recommended;
}).reduce(function(result, ruleData) {
    result[ruleData[0]] = 2;
    return result;
}, {})
```

PR #418, opened 2017-10-02
Perf improvements

denar90:
Hello, guys! 👋 
I've done some perf/PWA improvements.
Why they are useful:
 - perf for mobile devices are pretty critical recently 
 - personally, I'd like to have docs for OSS tools/libs PWA. I use public transport and it's free "LieFi"(c) which makes hard to read.surf docs going home/work.

TL;DR:

----

**urls**:
- http://localhost:4000/docs/user-guide/configuring ( because configs are always tough :) )
- http://localhost:4000/ (partly)

So, **conditions** of measurements
- Device Emulation (Nexus 5X)
- Network Throttling (150ms RTT, 1.6Mbps down, 0.7Mbps up)
- CPU Throttling (5x slowdown)

> all traces in dev tools were measured with fast 3G and 4x CPU throttling conditions

**Tools**:
 - [Lighthouse](https://github.com/GoogleChrome/lighthouse/)
 - [pwmetrics](https://github.com/paulirish/pwmetrics)
 - [timeline-viewer](https://chromedevtools.github.io/timeline-viewer/) for sharing traces

---

**Before**:
Lighthouse results:

image

pwmetrics results

image

> in case you are not familiar with metrics terminology - https://codeburst.io/performance-metrics-whats-this-all-about-1128461ad6b

**Timeline trace**:

https://chromedevtools.github.io/timeline-viewer/?loadTimelineFromURL=drive://0B0c67TI7mLzEZGwwblYxaVFqVjA

"We have some opportunity here" (c)

----

## Step 1. Get rid of jQuery.

jQuery is needed for `bootstrap` but we can use (bootstrap.native)[https://github.com/thednp/bootstrap.native] instead.

pwmetrics results:

image

**Timeline trace**:

https://chromedevtools.github.io/timeline-viewer/?loadTimelineFromURL=drive://0B0c67TI7mLzEZ1A5UEpkUFN5WWs

**Results**:
For FCP and FMP wins not so huge (100 ms) but we saved almost 1 sec (5,5sec => 4,6sec) for PSI.

---

## Step 2. Unblock FMP and FCP.
Looking below you can see that resources have either high priority which blocks FMP and FCP until they are loaded.

image

Let's add `defer` attribute to these and other scripts.

> In case someone is not aware of `defer` and `asyc` attributes - http://www.growingwiththeweb.com/2014/02/async-vs-defer-attributes.html

Another problem is some weird blinking of fonts.

image

In action

giphy 4

And move fonts loading to the body which is also "free" browser for parsing page without waiting for resource.

> For more info @jakearchibald has cool article - https://jakearchibald.com/2016/link-in-body/

pwmetrics results:

defer

FCP - 2,98 => 2,62 (300ms)
FMP - 2,98 => 2,62 (300ms)
PSI - 4,66 => 2,64 (2sec)

**Timeline trace**:
https://chromedevtools.github.io/timeline-viewer/?loadTimelineFromURL=drive://0B0c67TI7mLzELTNMZGxCWHYtNlU

---

## Step 3. Improve connection phase.

Using e.g. preconnect for google-analytics can save up to 100ms in a connection phase.

image

---

## Step 4. Improving thirdparty scripts.

`Anchor.js` is really useful tool, but it has some perf issues. We can't control loading styles/fonts because of tight integration of them in code

image

And screen of a trace:

image

So I created [PR](https://github.com/bryanbraun/anchorjs/pull/92) for this separation, but for now, we can have vendor js code here in repo.
This improvement helped us reduce on request for getting font because we can reuse `gylphicons` instead.

Final Lighthouse results
image

and pwmetrics results

improve-anchor

**Trace**
https://chromedevtools.github.io/timeline-viewer/?loadTimelineFromURL=drive://0B0c67TI7mLzEbzhIMmI0WU5SY0E

So comparing results **before and after**

image

So we end up with 1,7 seconds First Interactive win and 2,5 Perceptual Speed Index.

----

## Phase 2. PWA 🎉 

`eslint` docs use `https` so we are not blocked for service workers integration.
Changes like:

- Using [workbox](https://github.com/GoogleChrome/workbox) for service worker integration
-  `manifest.json` was added
- some images optimized a bit with [ImageOptim](https://github.com/ImageOptim/ImageOptim)

were added.

Aaaaaand latest tasty one:

(IntersectionObserver)[https://github.com/w3c/IntersectionObserver] !!!

The goal of using it - defer loading on main page images of companies.

In action

giphy 5

Lighthouse results:
image

## Conclusion:

- jQuery removed
- scripts/fonts were deferred 
- connection for google-analytics improved
- eliminated request for not needed fonts
- PWA

I think not so bad improvements to say that we reach PWA level.
Thanks :)

P.S. I'm not expecting fast merging of this one and open for any suggestions/improvements.


ilyavolodin:
Very nice PR. Thanks for doing that. I'm currently on vacation, but I'll review it in details next week. One thing that stands out from looking at it on the tablet is I don't think we want to touch logos. They are sort of properties of their respective owners, and I'd rather not mock with them.

denar90:
@ilyavolodin no worries, when you have a time :) 
>They are sort of properties of their respective owners, and I'd rather not mock with them.

I'm not sure I got your point correctly,  do you mean optimizing images with imageOptim? Wanna drop this commit?

_includes/footer.html
```diff
@@ -31,18 +31,37 @@
       <div class="suggestion-text">{{{text}}}</div>
     </div>
   </div>
-  {% endraw %}
+  {% endraw %}-
```
ilyavolodin: I think this `-` is extra

_includes/footer.html
```diff
@@ -31,18 +31,37 @@
...
   <script>
-    $(document).ready(function() {
+    document.addEventListener('DOMContentLoaded', function() {
+
+      if ("serviceWorker" in navigator) {
+        navigator.serviceWorker.register("/sw.js").catch(function(e) {
+          console.error("Error during service worker registration:", e);
+        });
+      } else {
+        console.log("service worker is not supported");
```
ilyavolodin: I think we can skip console logging. I'm not a fan of logging on prod websites. It's OK to log if there's some unexpected behavior, but in case service workers are not there, we should just skip it silently.

js/app/image-lazy-loader.js
```diff
@@ -0,0 +1,46 @@
+var ImageObserver = (function() {
+    var observer;
+
+    function createObserver() {
+
+        // small polyfill for Microsoft Edge 15 isIntersecting property
+        // see https://github.com/WICG/IntersectionObserver/issues/211#issuecomment-309144669
+        if ("IntersectionObserver" in window &&
```
ilyavolodin: IntersectionObservers are not supported on Safari on iOS or desktop. What's the fallback strategy for those browsers?

_includes/footer.html
```diff
@@ -31,18 +31,37 @@
...
+  {% endraw %}-
   </script>
-
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/1.10.2/jquery.min.js"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/3.3.4/js/bootstrap.min.js"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/anchor-js/1.1.1/anchor.min.js"></script>
-  <script src="{{ site.url }}/js/app/search.js"></script>
+  <script src="https://cdn.jsdelivr.net/g/lodash@3.10.1,hogan.js@3.0.2,algoliasearch@3.8.1,autocomplete.js@0.12.0(autocomplete.min.js)" defer></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/bootstrap.native/2.0.15/bootstrap-native.min.js" defer></script>
+  <script src="/js/app/anchor.min.js" defer></script>
```
ilyavolodin: I think the perf improvement for custom anchor is minimal, so I'd rather take a small hit in perf, but keep dependency intact. Unless you think this is critical.

package.json
```diff
@@ -12,7 +12,8 @@
     "start": "bundle exec jekyll serve --watch --incremental",
     "lint": "eslint js/app/ --ext=.js,.jsx",
     "test": "npm run lint",
-    "less": "lessc styles/less/main.less styles/main.css -x --source-map=styles/main.css.map"
+    "less": "lessc styles/less/main.less styles/main.css -x --source-map=styles/main.css.map",
+    "generate-sw": "workbox inject:manifest"
```
ilyavolodin: I haven't worked with PWA before, but if I understand this correctly, we will have to run this script on every release to regenerate checksums? Is that correct? If so, that would be a problem, since right now, we don't run any scripts on release.

_includes/footer.html
denar90: It causes around 700ms to load the additional font in case of using the original one. If you are ok with that then I'll revert this commit too.
image

package.json
denar90: To go the simplest way it's better to run `generate-sw` each time you release on prod. I'm not aware of a process so didn't add any changes to that.

_includes/footer.html
denar90: Sorry, I messed up, glyphicons it's own eslint site fonts. So it looks fine because of https://github.com/eslint/eslint.github.io/pull/418/files#diff-110b203ee900658cc6a1d3772a89c033R17
I'll use anchorjs together from cdn.

denar90:
Thanks, guys. I've added last changes. Always glad to help :)

denar90:
🎉  💃

PR #419, opened 2017-10-03
Updating migration guide for single-line disable

justinhelmer:
The multiline comment string doesn't seem to work.

Main docs page is right; JSCS migration guide appears to be wrong.


not-an-aardvark:
Thanks for the PR. Can you create this in the [`eslint/eslint`](https://github.com/eslint/eslint/blob/ee99876ccc76585b0af2c8673d7eb25bb0b972fa/docs/user-guide/migrating-from-jscs.md) repo instead? The file in this repo is autogenerated from `eslint/eslint`, so if you only modify it here it will get overwritten after the next release.

justinhelmer:
[no problem](https://github.com/eslint/eslint/pull/9385)

ilyavolodin:
I'm going to close this PR, since another one was filed in the ESLint repository. Thanks.

PR #421, opened 2017-10-10
Update eslint.org links to use https

not-an-aardvark:


platinumazure:
Why did you revert some of them?

not-an-aardvark:
@platinumazure I reverted modifying the formatter examples in https://github.com/eslint/eslint.github.io/commit/25922413e1cd2d5f5fccda3015b613c7d296a681 because those links are a description of ESLint's behavior. For example, in ESLint 3.0.0 the html formatter did output links with the `http` protocol, so it would be incorrect to show `https` in the example output.

Issue #422, opened 2017-10-11
Search bar fails with a type error


bwhitty:
![screen shot 2017-10-11 at 4 31 59 pm](https://user-images.githubusercontent.com/1571966/31472155-e60c2144-aea1-11e7-9831-c6eea1d1dad8.png)

This happens in both FF 57 and Chrome 61

ilyavolodin:
Thanks for the report. I'm looking into it now. It's related to PWA changes we pushed earlier today.

ilyavolodin:
This is now fixed (we reverted PWA until we can fix it). Closing

bwhitty:
Anyone who accessed the site while it's broken now has a zombie service worker installed for this site and so it continues to be broken today for those users.

<img width="1680" alt="screen shot 2017-10-16 at 12 50 29 pm" src="https://user-images.githubusercontent.com/1571966/31624432-af387852-b270-11e7-8512-e3f60d70aa0a.png">

Opening the site in an incognito window or manually deleting the cache files from "workbox-precaching" fixes this.

not-an-aardvark:
You can manually unregister it by entering this into the dev console:

```js
navigator.serviceWorker.getRegistrations().then(workers => workers.forEach(worker => worker.unregister()));
```

Obviously, we don't want every user to need to enter something into the dev console as a long-term strategy. Is there a good way to keep service workers up-to-date (aside from pushing an update that unregisters all of them)?

bwhitty:
I'm not sure of a solution off the top of my head. Here's my favorite talk about the subject, I vaguely remember it discussion this issue: https://www.youtube.com/watch?v=CPP9ew4Co0M

PR #428, opened 2017-10-14
Add service-worker 

denar90:
Let's do it one more time. After last PR were added some tune stuff and small bugfixes. You can try it live in my fork - https://denar90.github.io/eslint.github.io/

Diff between demo-page brunch and actual `perf-imps-try-2` branch - https://github.com/denar90/eslint.github.io/compare/perf-imps-try-2...denar90:gh-pages

Thank you for you time spent on me and be interested in perf :)


Aladdin-ADD:
image

seems this css file is not generated?

workbox-cli-config.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
+  "globDirectory": "./_site/",
```
ilyavolodin: Unfortunately this is not going to work. `_site` is a special folder generated by Jekyll, where it precompiles all of the pages. Unless you have Jekyll installed locally and run `jekyll build/serve/etc.` it will not be created. We run our site updates through the script executed on our CI server. Our CI server doesn't have Jekyll installed and because of that, this folder isn't going to be available. Using `/` isn't going to work either, because in the root, all of the files are partials, without headers and footers. They are also `md` files, and not `html` files.
In short, I really don't have a suggestion as to how to get this to work. The only thing I can think of, is installing Jekyll on our CI server, but I really don't want to do that. Jekyll is very brittle and breaks constantly because of the new versions released. Our CI server is also pretty small (hardware wise), and I really don't want to install Ruby on it, and slow it down even more.
@eslint/eslint-team anyone has any ideas on how we can do this?

denar90:
@Aladdin-ADD just some wrong path to file I think..

denar90:
I've tried to use runtime caching as @ilyavolodin mentioned. Also updated gh-pages for my fork.
https://denar90.github.io/eslint.github.io/
So we will cache pages which user has already seen. No precache for now...

denar90:
Guys, do you wanna go forward with this approach? Should I resolve conflicts?

ilyavolodin:
@denar90 Sorry about that. I've been extra busy this week with work and didn't have a whole lot of time for open source. I think we can go with the no precaching approach for now and see if in the future we would be able to tweak it to cache up-front. The only concern that I have with precaching is around cache invalidation. How would we handle that?

alberto:
Current performance is pretty good in my opinion, every page I tested loads under 1 second for me. I don't think the possible gain of adding sw is worth the hassle, for a static site like ours.

@denar90 is there any page where you think this is really needed? can you share some before/after measurements?

ilyavolodin:
@alberto This PR is not about performance. This is about making the site work offline, so that people can use documentation on their phones without connection.

denar90:
@alberto we've already merged part of of these changes in - https://github.com/eslint/eslint.github.io/pull/430
I suppose that's why you have so good results 😉 
So there is only stuff related to service workers.

denar90:
@ilyavolodin good point about cache invalidation. I think it should be handle it in scope of service worker version. We can think about it's updating e.g. after each deploy. Browser will that service-worker is outdated and install a new one.

alberto:
I mentioned performance because it was in the title and the description of this PR. 
I'm aware of all the other changes, and I thank you very much for those @denar90 

I'm questioning the need for SW caching. I haven't seen it discussed anywhere (maybe I missed another issue or PR?)

Is the offline support something that has been requested by a lot of people (or anyone actually)? I'm a bit reluctant to add it. Dealing with misconfiguration and cache issues like the one you mention can be difficult to detect and painful to fix. I think cache-first strategy can be problematic in this sense, where people get served their cached page instead of the new one after a release.

ilyavolodin:
@alberto Ultimately, we are already using cache-first strategy. Cloudflare works as a proxy for our site and caches all of the content. It's pretty smart about invalidating cache when needed, but we had issues once in a while with it not updating content on the site correctly.
To answer your question, yes, there were at least 3 requests that I can remember about making documentation available offline. At some point, we even discussed shipping documentation with the npm package and providing a CLI option for it.
@denar90 Would you be able to update this PR so we could look at your implementation of the ServiceWorker to see if we can spot any issues with cache invalidation?

alberto:
@ilyavolodin sorry I don't think that's a justification. If we have issues with cloudflare cache we should fix that on deploy: https://api.cloudflare.com/#zone-purge-all-files

Anyway, cloudflare honors the cache-control header, which is 5 minutes for html files. Service workers ignore that, so everyone willl experience the cache issue on every release.

I would like to hear some more opinions on this @eslint/eslint-team

not-an-aardvark:
I'm inclined to agree with @alberto about this, especially given that based on https://github.com/eslint/eslint.github.io/issues/422#issuecomment-336950760, it seems like if there is a buggy serviceworker it can make the site inaccessible even after the bug is fixed.

If someone wants to access the docs offline, couldn't they just clone this repository?

denar90:
Sorry guys this week extremely busy for me. 

@ilyavolodin I'll resolve conflicts asap.

@alberto if you are concerned about caching problems we can choose network first strategy which means that all resources will be fetched from network and in case user is offline it will grab resources from cache. Service worker will update the cache if new sw is registered, hence we just need to add a new version of it during each deploy which can be done automatically. 

@Aladdin-ADD I was dumb making mistake with replacing jquery by vanilla js (element.val() wasn't replaced by element.value) so you should be afraid of adding sw 😉 

In any case guys, trends are saying about more and more mobile usage of web apps. Personally, I'm using mobile for reading different kind of docs and think having pwa docs is good way going forward.

ilyavolodin:
@denar90 Not worries, no rush on this.
Btw, those trends are generic to the web. We are sort-of a special case, since we are tooling specific to desktop use, our mobile usage is actually really low. Only 1.7% of our traffic comes from mobile devices, and around 0.25% from tablets. However, there are advantages to using sw even on desktop, at least in the browsers that support it. We just need to be mindful of caching strategy.

alberto:
@ilyavolodin the actual number is smaller, considering iOS users won't benefit from that.

The caching strategy is one thing to consider, but also the potential impact of a bug in our code or the SW library. I am not sure that could be easily solved just with a deploy, so I doubt the potential benefits outweigh the risks

ilyavolodin:
Hmm.. If people think that it's not a good idea to add this, I will not insist. We already merged perf changes that I wanted to do for a while, so this is just an icing on top of the cake from my perspective. Although I was planning on adding ability to show notifications when the new version of ESLint is released to the SW later. @eslint/eslint-team what do others think about this?

denar90:
Sorry for the delay.
@alberto I've added [networkFirst](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-runtime-caching.NetworkFirst) strategy, so there should be no problems/concerns about cached resources. In other hand we open a new feature for users not breaking anything. Of course `workbox` (service worker library) might have issues, but it's well maintained so it will not be a problem.

It's clear that your users primarily use desktops, but what if by adding service workers number of them will grow on a mobile devices?
FYI, we will not add just offline feature but also [add to homescreen](https://developers.google.com/web/fundamentals/app-install-banners/) which works both desktop and mobile.

I know that other projects either add (e.g. https://twitter.com/XhmikosR/status/927276118566211584) or want to add service workers, besides, personally I'm a huge fan of pwa and perf. 

I propose to give it a try :)
P.S. In any case I'm grateful for your responses, spent time on this issue and open for adding changes if necessary.

ilyavolodin:
Thanks for doing this @denar90 @eslint/eslint-team How do you feel about this PR? I'm fine with having network first caching strategy. Worst case scenario, if we run into some problems, we can always push an update that will remove service worker.

denar90:
Guys, is something needed from my side to move it forward?

denar90:
cc @Aladdin-ADD @alberto @not-an-aardvark 
Can I bring your attention on this one? I think only one approve blocking merging this one, am I wrong? Do I have to add something?
Thanks :)

ilyavolodin:
Sorry, completely lost track of this issue. @eslint/eslint-issues-team anyone opposed to merging this in? Worst case scenario, if we don't like it, we can always add code later to unregister service worker.

not-an-aardvark:
> Worst case scenario, if we don't like it, we can always add code later to unregister service worker.

Is there a possibility that the code to unregister the service worker wouldn't end up being loaded, because it would be cached by the service worker? (I'm not very familiar with service workers in general.)

denar90:
@not-an-aardvark, I suppose no. If service worker unregistered then it can't handle http requests and work with cached data. So you should not be worried about this part :)

not-an-aardvark:
To clarify, I meant that we wouldn't be able to unregister the service worker because our code to unregister the worker wouldn't get loaded. I realize that there wouldn't be a problem after the service worker was unregistered. Is it possible that the service worker would inadvertently prevent itself from being unregistered?

denar90:
If we do something like this 
```
     navigator.serviceWorker.getRegistrations().then(function(registrations) {
        registrations.forEach(function(registration) {
          console.log('removing registration', registration);
          registration.unregister();
        })
      })
```

instead of https://github.com/eslint/eslint.github.io/pull/428/files#diff-7693eb643e8fe8b8d46b407909b408ceR46
then service worker will be definitely unregistered. Nothing like circular self caching like you was concerned:
>Is it possible that the service worker would inadvertently prevent itself from being unregistered?

not-an-aardvark:
Ok, sounds good to me.

alberto:
I don't want to sound like a broken record, but I think adding this can have a significant impact (much more than adding webpack which we have been quite conservative with) and at the same time be difficult to detect, so I think it should be a very conscious decision on our part and would ideally count with a champion in the team with significant experience.

Some sample gotchas people are dealing with:
https://redfin.engineering/service-workers-break-the-browsers-refresh-button-by-default-here-s-why-56f9417694
https://twitter.com/LeaVerou/status/931700832042700801
https://gist.github.com/Rich-Harris/fd6c3c73e6e707e312d7c5d7d0f3b2f9
https://twitter.com/AdamRackis/status/923600828073734144
https://twitter.com/samccone/status/922466599252123648

denar90:
I'm closing it, looks like no champion interested in it...

PR #430, opened 2017-10-15
Improvements without service workers

denar90:
Last PR #428 was divided and here present all that stuff except sw. I'm working on a problem related to deploy.


_includes/footer.html
```diff
@@ -31,18 +31,28 @@
       <div class="suggestion-text">{{{text}}}</div>
     </div>
   </div>
-  {% endraw %}
+  {% endraw %}-
```
ilyavolodin: This is from the older PR. Extra `-` symbol

denar90: thx

PR #433, opened 2017-10-24
Ensure demo works when running site locally

not-an-aardvark:
Currently, some resources are loaded from `https://eslint.org` even when running the site locally. Among other things, this was causing the demo to break because `requirejs` handled file extensions differently.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -112,7 +112,7 @@
 
   </script>
   {% if page.demopage %}
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.1.15/require.min.js" data-main="{{ site.url }}/js/app/demo/requireConfig" defer></script>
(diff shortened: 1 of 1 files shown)


denar90:
You should change site url in config, https://github.com/eslint/eslint.github.io/blob/master/_config.yml, than it will work fine. We should work around for solution running on dev and prod env.

not-an-aardvark:
Why is `site_url` needed here at all? It seems like this is what relative URLs are for.

denar90:
Sorry for the delay. It's not needed if you are working with local setup or main domain, but if you want to setup gh-pages or something then it fails. The point here if you want to drop site.url then you should drop it everywhere, otherwise we should workaround for automatic solution.

mysticatea:
LGTM.
This PR may resolve a problem https://deploy-preview-428--eslint.netlify.com/demo/ is not working.

PR #434, opened 2017-10-24
Handle demo config "opened" state in React (fixes #432)

### LINKED ISSUE #432: bug: demo configuration not shown
image

click configuration => not working.

not-an-aardvark:
This updates the "Configuration" bar to store its opened state in React, to ensure it opens correctly (fixes #432).

As a bonus, the entire bar is now clickable, rather than just the text.


christalk:
Thank you, it works now!

PR #445, opened 2018-01-12
Update: Add MongoDB to users

hswolff:


hswolff:
Ah! Apologies! Nice catch!

hswolff:
Fixed.

Issue #447, opened 2018-01-17
Improve "Getting Started" page?


platinumazure:
Original issue created as eslint/eslint#8420.

Quoting from the original issue:

> I am also hoping to see a better explanation of the [**Getting Started** article](http://eslint.org/docs/user-guide/getting-started), including:
> 
> - Explicitly say each rule do not support multiple options.
> - Explicitly say each rule has the potential to support unique severity beyond what we have.
> - A list of all options that each rule has. Currently, the options are written in the docs, but they are hard to distinguish against other similarly formatted text or code.
> - **Windows** users are being left out. Some of us, like me, are restricted to use Windows due to work. What is lacking is, for example, where the configuration files are at, and how to create a new `.eslintrc` manually without using `eslint --init` command (renaming a file as `.eslintrc.`, with 1 period at the beginning and end of the filename each).
> 
> I'm very certain there may be a duplicate issue, but the most important one is the point in boldfaced text.

I agree that the Getting Started section deserves another look. Or, maybe we should keep Getting Started very high-level and then point to more of a "cookbook" page that goes into some more detail about how to do some setup and tasks. We definitely need to give this some thought, at any rate.

mysticatea:
I agree that the Getting Started page needs improvement.

But I think "Windows users are being left out" is a misunderstanding. The `.eslintrc.` matter is not a restriction of Windows but a restriction of `explorer.exe`. Editors such as VSCode, Atom, or Sublime Text can create `.eslintrc` file without problem. I'm not sure that we should write about `.eslintrc` in `explorer.exe` because it has been deprecated. Currently `.eslintrc` needs an extension (`.js`, `.yml`, or `.json`).

platinumazure:
@mysticatea Yes, you're absolutely right. The fact that we don't mention the extensions shows how little care we've put into that section/page.

mysticatea:
Ah, oh, I got it. The Getting Started page has shown the deprecated feature...

platinumazure:
I made some improvements recently, but I'm not sure if we should be considering a new/different approach to how that page should present its information. I would love to hear what others think!

platinumazure:
I haven't heard anything from the rest of the team on this, and this hasn't been a priority for me either, so I'm going to go ahead and close this.

PR #448, opened 2018-01-17
Docs: Small updates to Getting Started page (refs #447)

platinumazure:
Incremental improvements to the Getting Started page. This page still might need to be redesigned a little bit, but at least some of the main issues have been addressed.

Question: Is there any point in mentioning the migration guides at all in a "Getting Started" page? If not, I'm happy to remove it outright.


docs/user-guide/getting-started.md
```diff
@@ -18,6 +18,17 @@ ESLint is a tool for identifying and reporting on patterns found in ECMAScript/J
...
 *Why ESLint* @0:00, *Installing and using ESLint* @2:20.  <a href="https://www.pluralsight.com/courses/eslint-better-code-quality?utm_source=eslint-dot-org&utm_medium=video&utm_campaign=authordemo" target="_blank">Full ESLint Course at Pluralsight</a>
 
+## Migrating from an Earlier Version
+
+If you are coming from an earlier version, please follow the migration guides below as needed:
+
+* [1.0.0](migrating-to-1.0.0)
+* [2.0.0](migrating-to-2.0.0)
+* [3.0.0](migrating-to-3.0.0)
+* [4.0.0](migrating-to-4.0.0)
```
j-f1: How about changing the links to something like

> 3.x → 4.x

?

mysticatea:
Migration doesn't sound "Getting Started" to me. I think that we can remove it from the page.

platinumazure:
@mysticatea @btmills Removed migration guide section. Fancy another review? Thanks!

PR #453, opened 2018-02-09
Update: Add background color for incorrect/correct examples

platinumazure:
I kept this very simple and just added a new .css page. Wasn't sure if I should integrate the CSS changes with our LESS setup or anything like that. I basically modeled this after the demo page's CSS.

I would appreciate feedback on color choices, especially from an accessibility perspective.


abiduzz420:
Looks good to me.

PR #456, opened 2018-03-15
Clarify default ignoring of node_modules

matijs:
When using ESLint from the root of a mono repo `node_modules` needs to be explicitly added to `.eslintignore` for ESLint to ignore for example `packages/*/node_modules`. Clarified this in the configuring documentation.

### DIFF
--- a/docs/user-guide/configuring.md
+++ b/docs/user-guide/configuring.md
@@ -840,15 +840,17 @@ Globs are matched using [node-ignore](https://github.com/kaelzhang/node-ignore),
 
 In addition to any patterns in a `.eslintignore` file, ESLint always ignores files in `/node_modules/*` and `/bower_components/*`.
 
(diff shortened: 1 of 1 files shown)


not-an-aardvark:
Thanks for the PR. This page is actually autogenerated from the [corresponding page](https://github.com/eslint/eslint/blob/72ed3dbcdda8ec1387e75676c3166fec71337a69/docs/user-guide/configuring.md) in the `eslint/eslint` repository; can you open a PR to that repo instead? (If we merge this PR, it will get overwritten the next time we do a release.)

matijs:
@not-an-aardvark oh that’s silly of me will do!

matijs:
Opened [a new PR](https://github.com/eslint/eslint/pull/10092) in the `eslint/eslint` repo.

PR #461, opened 2018-04-03
Upgrade algolia search plugin

ilyavolodin:
Hopefully, this will fix build issues and reindex the site.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,6 @@ branches:
   only:
     - master
 script: 
-  - bundle exec jekyll algolia push
+  - bundle exec jekyll algolia
 rvm:
  - 2.5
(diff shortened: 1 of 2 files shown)


platinumazure:
@ilyavolodin Do we need to update the netlify deploy metadata file (if there is one) and/or upgrade ruby for the same? Seems that there's an incompatibility that causes netlify to fail, but I don't know if that will cause our own gensite to fail. (See the netlify status check Details for more info)

ilyavolodin:
I'm just going to turn off netlify all together. It was added as a proof of concept, but it's not used by anyone and not needed.

not-an-aardvark:
The netlify build was generally useful when I was working on the demo, because it allowed team members to easily test changes in a PR without needing to clone it locally.

ilyavolodin:
Hmm.. Ok, I never used it, so I assumed that nobody else did either (I guess that's a bit presumptuous). We can restore netlify hook separately, but broken algolia search is affecting our core website, so I would rather try to fix that first, and then look into netlify issue.

PR #463, opened 2018-04-25
Update configuring.md

lukesalamone:
Remove comma

### DIFF
--- a/docs/2.0.0/user-guide/configuring.md
+++ b/docs/2.0.0/user-guide/configuring.md
@@ -41,7 +41,7 @@ Here's an example `.eslintrc.json` file:
         "sourceType": "module",
         "ecmaFeatures": {
             "jsx": true
-        },
+        }
     },
     "rules": {
         "semi": 2


ilyavolodin:
Thanks for PR @lukesalamone . Unfortunately, this file is auto-generated by our release scripts and the source is not located in the repository. The corresponding file that's used as a source is located here: https://github.com/eslint/eslint/blob/master/docs/user-guide/configuring.md Could you open a PR against that file in the main ESLint repository? Sorry about confusion.

platinumazure:
@ilyavolodin I don't think that's the case here. The changed file is in docs/**2.0.0**/user-guide/configuring.md (emphasis added). So this is really an archive and is not tied to eslint/eslint's docs/user-guide/configuring.md.

Let me know if I've missed something.

ilyavolodin:
Oh, sorry, that's my bad. I didn't notice that it was changing older version. In that case, everything is good.

PR #465, opened 2018-04-25
Fix: show popover for rulesConfig (fixes #437)

### LINKED ISSUE #437: Rule description popover doesn't work
When hovering over a rule name in the demo, a description of the rule is supposed to appear. However, that isn't happening right now, and an error is getting thrown:

```
Uncaught TypeError: $(...).popover is not a function
    at showPopover (https://eslint.orghttps://eslint.org/js/app/demo/rulesConfig.jsx:10)
    at Object.r (react.min.js:14)
    at i (react.min.js:12)
    at Object.u [as executeDispatchesInOrder] (react.min.js:12)
    at p (react.min.js:12)
    at f (react.min.js:12)
    at Array.forEach (<anonymous>)
    at r (react.min.js:16)
    at Object.processEventQueue (react.min.js:12)
    at r (react.min.js:14)
```

malcolmsgroves:
The bootstrap native popover objects were not being initialized. I added the initialization and refactored the show / hide toggle to match the bootstrap native [guide](https://thednp.github.io/bootstrap.native/)

### DIFF
--- a/js/app/demo/configuration.jsx
+++ b/js/app/demo/configuration.jsx
@@ -1,5 +1,7 @@
 "use strict";
 
+/* global Popover */
+
 define(["react", "jsx!parserOptions", "jsx!environments", "jsx!rulesConfig"], function(React, ParserOptions, Environments, RulesConfig) {
     return React.createClass({
         displayName: "Configuration",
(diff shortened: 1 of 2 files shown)


js/app/demo/configuration.jsx
```diff
@@ -6,10 +6,27 @@ define(["react", "jsx!parserOptions", "jsx!environments", "jsx!rulesConfig"], fu
         getInitialState: function() {
             return { showingConfig: false };
         },
+        initPopovers: function() {
+            var popoverElements = document.querySelectorAll("[data-toggle=\"popover\"]");
+
+            for (var i = 0; i < popoverElements.length; ++i) {
+                // eslint-disable-next-line no-new, no-undef
```
platinumazure: Why are you seeing `no-undef` errors here?

Is it due to the `Popover` reference? If so, the better way to handle that is to add Popover as a global, either to the main ESLint config or as a comment near the top of the file. See [configuration guide](https://eslint.org/docs/user-guide/configuring#specifying-globals) for more info.

PR #471, opened 2018-04-29
Fix: avoid changing URL fragment when switching demo tabs (fixes #469)

### LINKED ISSUE #469: Clicking on "fixed code" in the demo changes the URL
When clicking the "Fixed Code" tab on the demo, the URL fragment changes to `#fixedCode`, so the permalink to the code is lost.

Demo gif:

not-an-aardvark:
See https://github.com/eslint/eslint.github.io/issues/469 for details on the bug.

### DIFF
--- a/js/app/demo/app.jsx
+++ b/js/app/demo/app.jsx
@@ -110,10 +110,12 @@ define(["react", "jsx!editor", "jsx!messages", "jsx!fixedCode", "jsx!configurati
             }
             window.location.hash = Unicode.encodeToBase64(serializedState);
         },
-        enableFixMode: function() {
+        enableFixMode: function(event) {
+            event.preventDefault();
             this.setState({ fix: true });
         },
-        disableFixMode: function() {
+        disableFixMode: function(event) {
+            event.preventDefault();
             this.setState({ fix: false });
         },
         render: function App() {


js/app/demo/app.jsx
```diff
@@ -110,10 +110,12 @@ define(["react", "jsx!editor", "jsx!messages", "jsx!fixedCode", "jsx!configurati
             }
             window.location.hash = Unicode.encodeToBase64(serializedState);
         },
-        enableFixMode: function() {
+        enableFixMode: function(event) {
+            event.preventDefault();
```
not-an-aardvark: I'm not sure this is the ideal fix. I considered the following alternatives:

* Changing the tabs to use `<a>` tags without a `href` attribute: this is apparently not very good for accessibility, and it generated some linting problems from `eslint-plugin-jsx-a11y`.
* Changing the tabs to use `<button>` tags instead of `<a>` tags: This might be a better solution, but I had some trouble getting the styling to look the same when the tags are `<button>`s.
* Setting the `href` attribute for the tabs to the current URL: This would probably have worked but it doesn't seem any better than the `event.preventDefault` solution.

I decided to go with `event.preventDefault` since it seems like the simplest option.

PR #472, opened 2018-04-30
Chore: Run lint task in Travis

platinumazure:
This PR simply uses `nvm install` and then `npm run lint` to run the lint tasks. It also moves the algolia to an after_success step.

This is probably not the right approach. A better approach would be to see if we can switch the Travis build language to NodeJS and remove the algolia line entirely since we are now on DocSearch. This is just a POC (to try to see if it might be a viable temporary solution, should the DocSearch questions remain open for a while).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: ruby
 cache: bundler
-branches:
-  only:
-    - master
-script: 
-  - bundle exec jekyll algolia
+before_script:
+  - nvm install --lts
+script:
+  - npm run lint
+after_success:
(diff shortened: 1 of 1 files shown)


platinumazure:
Ack, I was really hoping this would trigger a Travis build. It must still be using the old logic (only master branch) despite the changes in this branch.

I'll dig in more later.

ilyavolodin:
No we don't need Ruby anymore. DocSearch doesn't require build step. So it should be fine to switch all builds to run normal Node image.

platinumazure:
Awesome, then I'll close this PR. Thanks!

PR #477, opened 2018-05-23
add smart polyfill

Aladdin-ADD:
Polyfill.io reads the User-Agent header of each request and returns polyfills that are suitable for the requesting browser. Tailor the response based on the features you're using in your app, and see our live examples to get started quickly.

refs: https://polyfill.io/v2/docs/

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -20,6 +20,7 @@
     </div><!-- /.container -->
 
   <link href="https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,600" rel="stylesheet">
+  <script src="https://cdn.polyfill.io/v2/polyfill.min.js" defer crossorigin></script>
   <script src="https://cdnjs.cloudflare.com/ajax/libs/bootstrap.native/2.0.15/bootstrap-native.min.js" defer></script>
   <script src="https://cdnjs.cloudflare.com/ajax/libs/anchor-js/4.1.0/anchor.min.js" defer></script>
     <!-- at the end of the BODY -->


platinumazure:
@eslint/eslint-team Is this something we need?

kaicataldo:
@Aladdin-ADD Do you mind explaining why this is needed?

Aladdin-ADD:
it can reduce the delivered js size for better perf, since just deliver the polyfills you need, tailored to each browser(based on UA).

for example, the latest chrome, the needed polyfill is 0. see https://polyfill.io/v2/docs/

ljharb:
I'd discourage doing this; first, polyfill.io's polyfills aren't always spec-compliant or complete, and second, using user-agent sniffing is unreliable.

All polyfills should always be unconditionally provided, for maximal correctness.

platinumazure:
@Aladdin-ADD What problem does this solve? Are users complaining about the size of the script, or performance loading the page, or anything like that? (And bear in mind, our online demo is not really a key part of what we need to support-- more important is the Node.js package.)

Is this something we truly need to do? At this point, I'm not convinced.

ilyavolodin:
We don't have any polyfills on the website. Our code is written to work on IE9 and above. I don't think we want to support older browsers anyways (we have no traffic from them, because our target audience are developers, we barely get any traffic at all from older browsers like IE11).

kaicataldo:
@ilyavolodin This would only be for the ESLint browser bundle that's generated for the demo (here's the [corresponding PR](https://github.com/eslint/eslint/pull/10642).

I don't think this is the right solve here. If we're going to make a change, I'd rather update how we use Babel and use babel-preset-env (with the [`useBuiltins` option](https://babeljs.io/docs/en/babel-preset-env#usebuiltins)).

PR #479, opened 2018-05-28
Fix: Add missing rules

platinumazure:
This will fix 404 responses for links that point to `max-classes-per-file` and `prefer-object-spread` rules.

Root cause is eslint/eslint-release#22. Hopefully we can fix that before next release!

### DIFF
--- a/docs/rules/max-classes-per-file.md
+++ b/docs/rules/max-classes-per-file.md
@@ -1,0 +1,65 @@
+---
+title: max-classes-per-file - Rules
+layout: doc
+edit_link: https://github.com/eslint/eslint/edit/master/docs/rules/max-classes-per-file.md
+---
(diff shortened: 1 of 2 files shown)


not-an-aardvark:
Since the files are being added manually, should we include a note at the top of the files that the rules are only available on an alpha release? It seems like someone could be surprised if they do `npm install eslint@latest` and some of the listed rules aren't present (although they could also check the bottom of the page to see when the rule was added, if necessary).

platinumazure:
@not-an-aardvark Agreed, but I would prefer to solve that problem separately/long term. This is just to fix broken links people are already reporting.

I created an issue on ESLint to discuss one possible approach: https://github.com/eslint/eslint/issues/10412

Issue #475, opened 2018-05-10
click search result, go to docs/3.0.0


Aladdin-ADD:
it is expected to open the latest version.

![image](https://user-images.githubusercontent.com/13050025/39847956-edf2f9ce-5436-11e8-849f-c313e62cd961.png)

ilyavolodin:
Yes, I noticed that too. It has something to do with this line: https://github.com/eslint/eslint.github.io/commit/6d951fdfb6ac8e28ba11d32d70b6117e7ac83e98#diff-7693eb643e8fe8b8d46b407909b408ceR31 but I haven't had a chance to dig into DocSearch documentation yet to figure out how to correct it.

Krinkle:
This is actively causing 404 Not Found errors for users, making it rather difficult to find things.

Example: search for "for-direction".

| Search | Result | 404
|--|--|--
|<img width="607" alt="screen shot of search query" src="https://user-images.githubusercontent.com/156867/41059069-75c7214e-69c3-11e8-933b-719890058d2c.png"> |<img width="685" alt="screen shot " src="https://user-images.githubusercontent.com/156867/41059112-925fea66-69c3-11e8-9394-8beea6472001.png">| <img width="798" alt="screen shot of 404 page with url" src="https://user-images.githubusercontent.com/156867/41059124-9a7db3fe-69c3-11e8-89da-e97f57f3fe59.png">

This claims to be the 3.0.0 rules page, but links to `for-direction` which was introduced in ESLint 4.

platinumazure:
I'm looking into this. I think I understand the issue, though I need to confirm a few things before I try to make any changes.

We'll also need to document necessary changes to docsearch configs (in https://github.com/algolia/docsearch-configs) as part of major releases. That will go in the maintainer guide.

platinumazure:
I've opened a PR on the docsearch-configs repo (algolia/docsearch-configs#440). The idea is to explicitly only index the "current version" of ESLint docs and not index old versions.

In that PR, I can't start from /docs because that page does not exist in the repo. So I'm starting from /docs/rules, /docs/user-guide, /docs/developer-guide, and /docs/maintainer-guide (all of which should exist). There might be other improvements we could make, but that PR (or something similar) should help fix most of the issues.

Note that we also need to make a change here in eslint.github.io (on the docsearch/autocomplete client-side configuration), so we would need to merge PRs in both repositories on the same day to avoid inconsistencies.

platinumazure:
[...]

PR #481, opened 2018-06-06
Update: Request docs and blog content in Docsearch

platinumazure:
This depends on algolia/docsearch-configs#440.

Refs #475. (If this and the docsearch-configs PR are merged, should fix that issue.)

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -28,7 +28,7 @@
     apiKey: '891b0e977d96c762a3821e0c00172ac9',
     indexName: 'eslint',
     inputSelector: '#eslint-search-input',
-    algoliaOptions: { 'facetFilters': [["version:3.0.0", "tags:blog"]] },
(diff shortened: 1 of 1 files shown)


s-pace:
👋 

These changes will reduce the scope of the search to the pages available under https://eslint.org/docs/**

We firstly need to merge the temporary index into the regular `eslint` one in order to make this changes work. I am waiting for your green light :) cc algolia/docsearch-configs#440

platinumazure:
Thanks @s-pace. My goal is for the search to return entries from /docs and /blog, did I mess up somewhere?

s-pace:
If you want the same behaviour as https://eslint-docsearch-475.surge.sh/, then you only need to remove the `algoliaOptions ` line from your script

Aladdin-ADD:
can I merge this? algolia/docsearch-configs#440 has been merged. @platinumazure

platinumazure:
Thanks, let's merge this.

PR #482, opened 2018-06-08
Fix: hidden icons on smartphone device

ybiquitous:
This fixes the problem that icons at the beginning of paragraphs are hidden on smartphone device.

`@media (max-width: @screen-xs-max)` means small devices (`< 768px`) in Bootstrap 3.
Please see <https://getbootstrap.com/docs/3.3/css/#grid-media-queries>.

### Before

image

### After

image


platinumazure:
Merging, thanks for contributing!

PR #484, opened 2018-06-09
Docs: Fix blog post for v5.0.0-rc.0 release

platinumazure:


platinumazure:
Oops, maybe this should have waited a day since the DocSearch crawler probably hasn't run since the docsearch-configs PR was merged. So now our search has a bad configuration until the index is updated.

I don't think it's worth reverting at this point, though. It should fix itself within 24 hours.

PR #487, opened 2018-06-23
Update blogpost for v5.0.0

not-an-aardvark:


_posts/2018-06-22-eslint-v5.0.0-released.md
```diff
@@ -7,11 +7,21 @@ tags:
 ---
 # ESLint v5.0.0 released
 
-We just pushed ESLint v5.0.0, which is a minor release upgrade of ESLint. This release adds some new features and fixes several bugs found in the previous release.
```
not-an-aardvark: It seems a bit odd that the release tooling classified this as a "minor" release (this happened because there were no breaking changes since the last prerelease, but that's probably not the comparison that most readers would expect). Maybe we should fix that for next time.

platinumazure: Possibly related: https://github.com/eslint/eslint-release/issues/23

PR #492, opened 2018-06-30
Fix: add missing link to "custom parsers" page

g-plane:


Aladdin-ADD:
thanks for the contributing!

PR #497, opened 2018-07-12
Linkify npm link in portmortem

btmills:


btmills:
Nevermind, @hzoo took care of it in #496

PR #506, opened 2018-10-03
Update: Use official ecmaVersion names in demo

not-an-aardvark:
This updates the parser options in the demo to use values like "2015" rather than "6" for the ecmaVersion option.

### DIFF
--- a/js/app/demo/parserOptions.jsx
+++ b/js/app/demo/parserOptions.jsx
@@ -7,7 +7,7 @@ define(["react"], function(React) {
                 <div className="col-md-4">
                     <h3>ECMA Version</h3>
                     <select
-                        value={props.options.ecmaVersion + (props.options.ecmaVersion < 2015 ? 2009 : 0)}
+                        value={props.options.ecmaVersion}
                         onChange={
                             function(event) {
                                 props.onUpdate(Object.assign({}, props.options, { ecmaVersion: +event.target.value }));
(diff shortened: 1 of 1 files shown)


js/app/demo/parserOptions.jsx
```diff
@@ -7,7 +7,7 @@ define(["react"], function(React) {
                 <div className="col-md-4">
                     <h3>ECMA Version</h3>
                     <select
-                        value={props.options.ecmaVersion}
+                        value={props.options.ecmaVersion + (props.options.ecmaVersion < 2015 ? 2009 : 0)}
```
j-f1: What would this do if you selected 3 or 5?

Aladdin-ADD: good catch! I believe  this change is not necessary, since we have normalized the config in eslint. https://github.com/eslint/eslint/blob/066f7e07d4f722573c6dbe7f64b0765504595644/lib/linter.js#L370-L387

not-an-aardvark: @j-f1 Good catch, will fix that case.

@Aladdin-ADD I think it's useful to display the official ecmaVersion to the user in the UI, but I agree that it's probably not necessary to change the internal representation.

Aladdin-ADD: agreed! I just mean the change of this line is not necessary. sorry for confusing! :(

PR #507, opened 2018-10-03
Update: remove experimentalObjectRestSpread option from demo

not-an-aardvark:
The `experimentalObjectRestSpread` option is deprecated, so it's probably not necessary to have it in the demo.

### DIFF
--- a/js/app/demo/app.jsx
+++ b/js/app/demo/app.jsx
@@ -89,6 +89,18 @@ define(["react", "jsx!editor", "jsx!messages", "jsx!fixedCode", "jsx!configurati
                 }
             );
 
+            if (
+                initialState &&
+                initialState.options &&
+                initialState.options.parserOptions &&
+                initialState.options.parserOptions.ecmaFeatures &&
+                initialState.options.parserOptions.ecmaFeatures.experimentalObjectRestSpread
+            ) {
+
+                // Delete the `experimentalObjectRestSpread` option from old states created when the demo supported the option.
+                delete initialState.options.parserOptions.ecmaFeatures.experimentalObjectRestSpread;
+            }
+
             this.initialText = initialState.text;
             return initialState;
         },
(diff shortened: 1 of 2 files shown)


js/app/demo/app.jsx
```diff
@@ -89,6 +89,18 @@ define(["react", "jsx!editor", "jsx!messages", "jsx!fixedCode", "jsx!configurati
...
+            if (
+                initialState &&
+                initialState.options &&
+                initialState.options.parserOptions &&
+                initialState.options.parserOptions.ecmaFeatures &&
+                initialState.options.parserOptions.ecmaFeatures.experimentalObjectRestSpread
+            ) {
+
+                // Delete the `experimentalObjectRestSpread` option from old states created when the demo supported the option.
+                delete initialState.options.parserOptions.ecmaFeatures.experimentalObjectRestSpread;
```
Aladdin-ADD: is it just working for the demos previously created, but not working for newly created?

not-an-aardvark: With the new version of the demo, it won't be possible to configure `experimentalObjectRestSpread` (without manually changing the URL hash), so this will generally only affect existing URLs.

not-an-aardvark: The demo always uses the latest version of ESLint, so the results frequently change. (For example, if someone reports a bug using the demo, and then the bug is fixed, the original link in the demo will no longer reproduce the bug.) Demo URLSs using `experimentalObjectRestSpread would end up breaking in ESLint v6 when we remove the option, regardless of what we do here.

With this change, I'm just trying to prevent a scenario where the option is still *active* and affecting the parse results despite not being displayed in the UI, because this seems like it would make the behavior very confusing.

js/app/demo/app.jsx
```diff
@@ -89,6 +89,18 @@ define(["react", "jsx!editor", "jsx!messages", "jsx!fixedCode", "jsx!configurati
                 }
             );
 
+            if (
+                initialState &&
+                initialState.options &&
+                initialState.options.parserOptions &&
+                initialState.options.parserOptions.ecmaFeatures &&
+                initialState.options.parserOptions.ecmaFeatures.experimentalObjectRestSpread
```
kaicataldo: 😄

PR #511, opened 2018-10-08
Update command-line-interace.md

henryemekaonuorah92:
Fix for Documentation about Caching is outdated #10739


platinumazure:
Hi @henryemekaonuorah92, thanks for the PR.

The original issue was fixed by eslint/eslint#10913, and the live site will be updated with our next release. Our next release will be this Friday or Saturday (12-13 October). So I'm not 100% sure this change is needed right now (and if we did want to put it in early, it should have the new message from the other PR).

Additionally, we would need you to sign our Contributor License Agreement.

Please let me know if you want to put in the other text from eslint/eslint#10913 and sign the CLA, and we can consider this. Otherwise, this is resolved in ESLint and just needs to be released. Thanks!

not-an-aardvark:
I'm going to close this PR because the fix from https://github.com/eslint/eslint/pull/10913 has been applied to the live site. Thanks for the contribution regardless!

PR #516, opened 2018-10-26
Fix sort-imports example

anoadragon453:
Correct and incorrect code examples were the same, which leads me to believe that this is what the correct code example should be.


platinumazure:
Hi @anoadragon453, thanks for the PR!

Would you mind opening this on the [eslint](https://github.com/eslint/eslint) repository? We could merge this to fix the website now, but it will be overwritten on our next release. If we do this fix in the main repository, then it won't fix the site immediately but it will fix the site on the next release.

Let me know if that didn't make sense. :smile:

anoadragon453:
Sure, no worries, will get right on that. Thanks!

ilyavolodin:
Closing this PR since the issue is in the main ESLint repository.

PR #523, opened 2018-11-15
Add icon color for emphasize rule is correct or not

kewang:
In order to emphasize rule is **correct** or **incorrect**, I modify the `thumb-*` icon color

### DIFF
--- a/styles/less/overrides.less
+++ b/styles/less/overrides.less
@@ -62,13 +62,20 @@ p {
 
   &.icon {
     position:relative;
+
+    &.correct:before {
+      color: #228b22; /* use green color stand for *correct* */
+    }
+
+    &.incorrect:before {
+      color: #dc143c; /* use red color stand for *incorrect* */
+    }
   }
 
   &.icon:before {
     position: absolute;
     left: -1.5em; /* negative padding-left of ul */
     font-family: "Glyphicons Halflings";
-    color: #4d4d4d; /* hsl(0,0%,30%) is 10% lighter gray than text */
 
     @media (max-width: @screen-xs-max) {
       position: static;


kaicataldo:
Thanks for the PR! Do you mind including some screenshots of your change?

kewang:
Like this.

kaicataldo:
I like the idea of this change. My only concern is that green/red isn't great for usability for color blindness. We do have the different symbols to differentiate, but I wonder if we could pick some colors that are friendlier for visual disabilities.

kewang:
The green/red color is common sense for identify yes/no globally. I have no idea if we pick another colors, I am just an engineer only. Orz

kaicataldo:
Yes, it's definitely commonly used. Just wondering if there's another option here that would allow us to be more inclusive of those with visual disabilities.

j-f1:
How about ✅ and ❌ icons instead?

kewang:
[Visual Impairments - Color Vision @ w3c](https://www.w3.org/TR/low-vision-needs/#color-vision)

👍👎⭕❌✔️ are good choices, but they have the same color problem like @kaicataldo said comment.

I think it is a hard problem for me.

kaicataldo:
Just want to be conscientious of this if we can. I don't want to be a blocker, though, if others are fine with this as is.

kewang:
I think maybe we can review the accessibility issue at whole site. Of course, it is another issue.

kaicataldo:
@kewang That's a good idea!

alberto:
I'm with @not-an-aardvark. Since color isn't the only differentiator but an extra visual reinforcement, I don't think this is an accessibility issue.

PR #526, opened 2018-11-30
New: Add /team page

nzakas:
I've added a `/team` page an updated the "About" link in the heading to be a dropdown that has "About ESLint" and "About the Team". The grid is responsive, going down to just two columns on mobile and up to four columns on larger screens.

screen shot 2018-11-30 at 11 06 33

### DIFF
--- a/_data/team.json
+++ b/_data/team.json
@@ -1,0 +1,158 @@
+{
+    "tsc": [
+        {
+            "username": "nzakas",
+            "name": "Nicholas C. Zakas",
+            "website": "https://humanwhocodes.com",
+            "avatar_url": "https://avatars3.githubusercontent.com/u/38546?v=4"
+        },
+        {
+            "username": "platinumazure",
+            "name": "Kevin Partington",
+            "website": "",
+            "avatar_url": "https://avatars0.githubusercontent.com/u/284282?v=4"
+        },
+        {
+            "username": "ilyavolodin",
+            "name": "Ilya Volodin",
+            "website": "",
+            "avatar_url": "https://avatars1.githubusercontent.com/u/1380062?v=4"
+        },
+        {
+            "username": "btmills",
+            "name": "Brandon Mills",
+            "website": "https://bmills.net",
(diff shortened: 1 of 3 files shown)


nzakas:
Just a note about the data: I wrote a utility to create `_data/team.json` directly from the GitHub API. We can just run that as part of the build to update it automatically. It pulls from the GitHub team data.

And no, I didn't edit the data to put me first. That's the way it came out of GitHub. :)

not-an-aardvark:
re. eslintbot, it would be simpler to remove it from `eslint-team`. I've just done that now, so if the list is regenerated the bot will no longer appear.

It needs admin access to repositories in order to push releases, but it should already have that due to being an organization owner.

---

As a sidenote, I've was looking into whether it would be possible to get rid of the @eslintbot account entirely and only use the `eslint[bot]` GitHub app for releases. It seems like it's currently not possible because the bot needs to push release commits to a protected branch while overriding the requirement for PR approvals/status checks. This can only currently be done by real accounts (not GitHub apps) with admin access to the repo. (See [here](https://platform.github.community/t/repositories-which-have-protected-branches-with-push-restrictions-have-no-ability-to-grant-push-rights-to-integrations/1376) for more info.)

_data/team.json
```diff
@@ -0,0 +1,158 @@
...
+    ],
+    "committers": [
+        {
+            "username": "Aladdin-ADD",
+            "name": "薛定谔的猫",
+            "website": "",
+            "avatar_url": "https://avatars2.githubusercontent.com/u/13050025?v=4"
+        },
+        {
+            "username": "eslintbot",
```
not-an-aardvark: I'm making a "request changes" review as a reminder to regenerate this list, which will have the effect of removing `eslintbot`. Feel free to dismiss this review and merge after `eslintbot` has been removed.

nzakas:
I re-ran the utility and updated `team.json`. I'll work to get that script integrated into the `gensite` build on the main repo.

PR #527, opened 2018-12-04
Build: Add script to generate team data (fixes #515)

### LINKED ISSUE #515: Create a /team page
Create a page at /team that lists out the ESLint team. This should have three sections:

* Technical Steering Committee
* Committers
* Alumni

I envision each section as a heading with images/names/GitHub usernames, similar to the [Babel page](https://babeljs.io/team). Unlike the Babel page, I don't think having a Twitter username is useful, but open to other opinions on that.

I'm pretty sure this can be built automatically using the GitHub API and a script, so that would be my approach.

nzakas:
This PR adds a script that generates the data in `_data/team.json`. The plan is to have this run periodically using a Jenkins build.


_tools/fetch-team-data.js
```diff
@@ -0,0 +1,87 @@
+/**
+ * @fileoverview Script to fetch team data from GitHub. Call using:
+ *     node fetch-team-data.js
```
not-an-aardvark: Based on the relative filepath of `"./_data/team.json"` below, it seems like this script is intended to have the project root as the CWD. However, these usage instructions seem to imply that the script should have `_tools` as the CWD so that the script would be found, which could potentially be confusing.

```suggestion
 *     node _tools/fetch-team-data.js
```

nzakas:
Good catch.

_tools/fetch-team-data.js
```diff
@@ -0,0 +1,87 @@
...
+
+// final data structure
+const team = {
+    tsc: [],
+    alumni: [],
+    committers: []
+};
+
+//-----------------------------------------------------------------------------
+// Requirements
```
j-f1: Why does this say “Requirements” again?

nzakas: Oops, just a copy paste error.

nzakas: Oops, just a copy-paste error.

```suggestion
// Main
```

PR #538, opened 2019-02-11
New: Incorporate sponsors into the site

nzakas:
* Move "Who's using ESLint" to `/users`
* Add "Who's using ESLint" to About menu
* Add script to pull backers from Open Collective
* Incorporate sponsors on the homepage
* Add "Become a Sponsor" at the bottom of blog post layout

Once this is merged, I can setup a Jenkins job to automatically pull new sponsors periodically so we don't have to manually update the site.


ilyavolodin:
Could you post a screenshot of the homepage? Also, there seems to be an issue with OpenCollective feed, It's marking one donation as 2 separate people/orgs because it was done through gift card.

nzakas:
The backer data comes directly from Open Collective, so if there's a bug, it's on their end (the same info is displayed at the top of the Open Collective page). I can ping Pia to see what she thinks.

Here's what the homepage looks like:

screen shot 2019-02-11 at 09 10 17

ilyavolodin:
Yes, looks like OpenCollective bug. We should probably report it.

PR #551, opened 2019-03-07
Build: switch to webpack

g-plane:
Switching to webpack, it allows us to use ES6 `import` in source file.


js/app/.eslintrc.yml
```diff
@@ -13,6 +13,7 @@ parserOptions:
     ecmaVersion: 5
     ecmaFeatures:
         jsx: true
+    sourceType: module
```
aladdin-add: I don't think we can use module here, given ecmaVersion is setting to 5.

webpack.config.js
```diff
@@ -0,0 +1,10 @@
+const path = require("path");
+
+module.exports = {
+    mode: "production",
+    entry: "./js/app/parser/index.js",
+    output: {
+        filename: "index.built.js",
+        path: path.resolve("./js/app/parser")
```
aladdin-add: can you use `path.join()` here,  I remember we encountered a similar issue in eslint repo somehow.

js/app/.eslintrc.yml
g-plane: So should we revert it to use CommonJS instead of ES Modules?

aladdin-add: I guess so.…（⊙＿⊙；）…
 another way would be adding something like babel, but it needs to be discussed more.

webpack.config.js
g-plane: The `path` field must be absolute path, while using `path.join` won't return that.

kaicataldo: I think this is fine, though I do think it's common to use `__dirname`. i.e. `path.resolve(__dirname, "js/app/parser")`.

js/app/.eslintrc.yml
kaicataldo: I agree - I don't think we should necessarily update to use ES modules here. That seems like a bigger discussion around if we want to use other ES2015+ features or not. I've been away for a bit (so please correce me if I'm wrong!), but I don't think we actually have a real automated build step at the moment.

g-plane:
I'm closing now since I think current build steps work fine.

PR #552, opened 2019-03-12
docs: fix JSON Schema example dead link

kazupon:
I found a dead link of JSON Schema example.
In the JSON Schema official site, I couldn't find the equivalent page.
For this reason, I've fixed to be similar documentation.


not-an-aardvark:
Thanks for the pull request!

This page is autogenerated from the [corresponding page in the `eslint/eslint` repository](https://github.com/eslint/eslint/blob/e7266c2478aff5d66e7859313feb49e3a129f85e/docs/developer-guide/working-with-rules.md), so these changes would get overwritten next time we build the site. May you please submit this pull request to the `eslint/eslint` repository instead?

kazupon:
Thank for your quick reply!
I got it!

PR #562, opened 2019-05-05
Added  ES2019 to demo

AlokTakshak:


AlokTakshak:
> could you please add it in the parser demo, or put it in a shared config?

can you specify which parser demo or shared config?

aladdin-add:
emm, seems it has been added:

https://github.com/eslint/eslint.github.io/blob/bff9433e3797fe5ae0cdbd20729619b0d9caa98b/js/app/parser/parser.config.js#L11-L18

PR #566, opened 2019-05-08
Docs: updated blockquote css

AlokTakshak:


AlokTakshak:
> Let's revert the CSS changes for now and see how things look with just the blockquote removal in the Markdown source. If we need to add a CSS change later, we can do so. Thanks!

image

btmills:
@AlokTakshak apologies for not seeing this before, and thanks for the tag @platinumazure. When we do a release of ESLint, the build process regenerates the eslint.org site in this repository from documentation stored in eslint/eslint, which is the authoritative copy. Here’s this file over in the main repository: https://github.com/eslint/eslint/blob/master/docs/developer-guide/working-with-rules.md Can you open a pull request in the eslint/eslint repository with this change and tag me or request a review so I get a notification? That’ll keep your fix from being overwritten by the next release.

It looks like the blank lines before `countOptions` And `rangeOptions` were lost in this change. Those prevent Markdown from joining those lines onto the previous list item.

Thanks for your patience in getting this fixed!

AlokTakshak:
@btmills  i tagged you in the other PR on `eslint` and here also i have updated the line space

btmills:
This was fixed in eslint/eslint#11818, and it's now visible in the [v6 prerelease docs](https://eslint.org/docs/6.0.0/developer-guide/working-with-rules#contextgetsourcecode)! The fix will make its way into the main docs once v6.0.0 is released. Thanks for fixing this @AlokTakshak!

PR #573, opened 2019-06-10
Docs: make a URL a link.

mysticatea:
I have noticed the URL was not a link.


platinumazure:
Thanks @mysticatea, I missed this.

Issue #490, opened 2018-06-24
travis-ci build no longer triggered somehow


aladdin-add:
build failing, and it has not been triggered for a long time. see https://travis-ci.org/eslint/eslint.github.io

PR #152, opened 2015-11-06
Add Algolia search

pixelastic:
As discussed in https://github.com/eslint/eslint.github.io/issues/69, here is my proposal for adding search to the eslint website using the Algolia API. You can test it live on http://eslint.algolia.com/.

2015-11-06_17h33m35

It was done using our [Jekyll plugin](https://github.com/algolia/algoliasearch-jekyll). It includes a few specific tweaks to better adapt to the eslint doc (does not search in the blog changelogs, nor in previous rule versions for example). I've added several RWD displays of the searchbar + results.

It is currently hosted on one of my demo accounts, but if you open an account on Algolia, I'll make it unlimited and move all the data to it.

I've also added the necessary travis config files so it should rebuild the search index on every new push on master. It only requires you to configure your Travis with the correct `ALGOLIA_API_KEY` environment variable ([more info here](https://github.com/algolia/algoliasearch-jekyll#3-configure-travis)). Otherwise, you just have to call `jekyll algolia push` to re-index manually.

Tell me what you think

PR #578, opened 2019-06-17
Chore: clean up dead code after DocSearch switch

kaicataldo:
fixes https://github.com/eslint/eslint.github.io/issues/490

Note: I turned Travis CI on for PRs to test this out. I think we should leave it on, but please let me know if you disagree with me.

I started digging into why CI has been failing, and it appears to me to be because the script we're trying to run in Travis CI is `bundle exec jekyll algolia`, however, we [removed the algolia gem](https://github.com/eslint/eslint.github.io/pull/464) last year.

[This commit](https://github.com/eslint/eslint.github.io/commit/ffa14ce38f9221bf45644907d4bf4c30dec47911) is the first commit to start consistently failing CI, and it looks like that's the first commit to be run with Travis (in other words, Travis has been failing since we turned it on).

I believe all the deletions here are no longer used. The only I'm not 100% sure about is `_plugins/search.rb`, but it looks like we're using Algolia's DocSearch API which runs entirely in JS and that we don't use this plugin any more. I've run the build locally and, as far as I can tell, search continues to function normally.

If/once we merge this, I'll also remove the `ALGOLIA_API_KEY` environment variable in Travis, since we it doesn't seem like that is used either any more.

Please let me know if I'm missing something important and deleting something I shouldn't be!


.gitignore
```diff
@@ -1,6 +1,5 @@
 node_modules/
 _site/
-_algolia_api_key
```
kaicataldo: As far as I can tell, this file never existed. See https://github.com/eslint/eslint.github.io/pull/152#issuecomment-155046398 for more details, but here's the relevant bit:

> If you want to push data from your local machine, you just have to use ALGOLIA_API_KEY=xxx jekyll algolia push. On Travis, you just have to declare this variable as an env variable in your config. You can also create a file _algolia_api_key that contains the key, at the root of your jekyll repo (I've ignored it in the .gitignore, just to be safe)

.travis.yml
```diff
@@ -1,9 +1,9 @@
...
 cache: bundler
 branches:
   only:
     - master
-script: 
-  - bundle exec jekyll algolia
-rvm:
- - 2.5
+script:
+  - bundle exec jekyll build
```
kaicataldo: This is the [recommend way](https://jekyllrb.com/docs/continuous-integration/travis-ci/) to test in CI that the site will build correctly

styles/less/search.less
```diff
@@ -45,9 +45,6 @@
 	.eslint-navbar {
 		clear: both;
 	}
-	.algolia-autocomplete {
```
kaicataldo: I don't see any references to this class in our templates nor when I search the DOM. I believe this is no longer used.

ilyavolodin:
Yes, your guesses were correct. Once we removed Algolia gem, build was no longer needed at all. I just never got around to removing it.

kaicataldo:
In addition to merging this, I have also deleted the `ALGOLIA_API_KEY` environment variable in the Travis build.

PR #582, opened 2019-06-22
Fix: Demo now uses updated Linter access point (fixes eslint/eslint#11873)

platinumazure:
This commit just fixes the access point to get the demo working. Might not be the best long-term approach.

I'm also exploring (separate from this PR) whether it might be possible to add some unit tests to at least test that the React component can be generated.


aladdin-add:
I don't think we should use `lib/linter.js`, since it is something non-public, and may changed in future versions. just created a PR(https://github.com/eslint/eslint/pull/11877) to fix it.

kaicataldo:
The deploy preview feature with Netlify is very cool

mysticatea:
Thank you so much!

PR #583, opened 2019-06-23
Fix: fix site name in title

g-plane:
The site name is missing at `<title>` of the pages, and this PR fixes it.

Page template: https://github.com/eslint/eslint.github.io/blob/26276d3554706862f6a280f20063ab7e74108b81/_includes/header.html#L24

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,4 +1,4 @@
-name: "ESLint - Pluggable JavaScript linter"
+title: "ESLint - Pluggable JavaScript linter"
 repository: "eslint/eslint.github.io"
 exclude: [
   'node_modules',
--- a/_config.yml
+++ b/_config.yml
@@ -1,4 +1,4 @@
-title: "ESLint - Pluggable JavaScript linter"
+name: "ESLint - Pluggable JavaScript linter"
 repository: "eslint/eslint.github.io"
 exclude: [
   'node_modules',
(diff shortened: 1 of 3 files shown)


kaicataldo:
I actually changed this in a recent PR. I didn't realize this was actually being used in the templates - sorry about that!

The Jekyll documentation doesn't actually have `site.name` or `site.title` listed (only `page.name` and `page.title`), and it looks like there's confusion in the community about which to use (See https://github.com/jekyll/jekyll-feed/issues/47 and https://github.com/jekyll/jekyll-seo-tag/issues/238). In the issues listed above, it seems like the community is moving toward using `site.title`.

Additionally, in the [Jekyll documentation](https://jekyllrb.com/docs/variables/#page-variables), it says the following about the corresponding `page` variables:
* `page.name` - "The filename of the post or page, e.g. about.md"
* `page.title` - "The title of the Page."

Extrapolating from that, I do think it makes sense to have `site.title` be the title of the page, and to instead update the header template to have `<title>{{ page.title }} - {{ site.title }}</title>`. What do you think?

g-plane:
Yes, `site.title` is recommended by Jekyll from the build output of Netlify. However, I may update it tomorrow.

kaicataldo:
For posterity, here's the error output from Jekyll (and what originally alerted me to the discrepancy when I made the change): 

`GitHub Metadata: site.name is set in _config.yml, but many plugins and themes expect site.title to be used instead. To avoid potential inconsistency, Jekyll GitHub Metadata will not set site.title to the repository's name.`

kaicataldo:
Thanks for contributing!

Issue #576, opened 2019-06-14
Proposal for adding a build step


kaicataldo:
Originally from https://github.com/eslint/eslint.github.io/issues/536#issuecomment-500599653.

In pursuing https://github.com/eslint/eslint.github.io/issues/536, I would also like to revisit the idea of adding a build step to our site. We know it would be a prerequisite for any of the options we've discussed in https://github.com/eslint/eslint.github.io/issues/536, and separating that task out into its own chunk of work feels like a much easier/cleaner way to do things. Besides unblocking us from using a different static site framework, I believe we would get the following immediate benefits:

1. When I [first proposed this](https://github.com/eslint/eslint.github.io/pull/405), we weren't pulling data from Open Collective's API. The build step could now automate running less, bundling the parser and demo, and running the script we use to pull data from Open Collective. Currently these are run by individuals before committing, which is extra overhead and as a process is error prone (someone might forget to run the script). Automating it would ensure it happens on every commit. **Edit: @ilyavolodin corrected me below to say that the Open Collective scripts are actually run on the hour and should be included in this.**
1. We would get to dogfood ESLint in a more modern web application.
1. I believe this would make it more approachable and enjoyable for our community to contribute to the site (particularly the demo).

Here is how I propose we move to having a build step:

1. We could use the current repository, but GitHub pages has the unfortunate limitation of only allowing hosting from the `master` branch for a user or org site (one that ends in `*.github.io`). We could create a separate `src` branch, but that feels confusing and would require anyone working on the site to know that `master` is essentially a build artifact. If we were to create a new repository called `eslint/website`, we could treat the `master` branch as the source and publish to a second branch called `gh-pages`. This way, `master` is always the source of truth.
1. If we do create a new repository, we would have to update the release script to also copy documentation files into that repo in the interim.
1. We set up Travis (as described [here](https://docusaurus.io/docs/en/publishing#using-travis-ci)) to run whenever a commit lands in master, running the build and publishing it to the `gh-pages` branch.
1. Update the build script to bundle the demo and the parser with Webpack, run less, and run the script that pulls from the Open Collective API.
1. At this point, we can evaluate how this is working and perhaps take it to a TSC vote to determine if we want to move forward with a build step.
1. Once we cut over completely to that repository for the website, we would have to update the release script once more to only generate docs in the new repository.
1. Deprecate the current repository in favor of the new one.
[...]

Issue #543, opened 2019-02-18
discussion: what sort of minimum versions we need to support?


aladdin-add:
refs:https://github.com/eslint/eslint/pull/11398#discussion_r257776506
a bit related to #537

j-f1:
@jamiebuilds’s [suggestion](https://jamie.build/last-2-versions) of `>0.25%, not ie 11, not op_mini all` is one option.

g-plane:
We can consider creating a "modern build" like `vue-cli` does.

aladdin-add:
> @jamiebuilds’s suggestion of >0.25%, not ie 11, not op_mini all is one option.

at the very least, we still need to support IE11, for window 7 users -- Edge requires window 10. But I'd like to see someone share the GA data.

+ https://support.microsoft.com/en-us/help/11531/internet-explorer-system-requirements
+ https://www.microsoft.com/en-us/download/details.aspx?id=48126

g-plane:
However, the ESLint website itself doesn't use Babel currently.

ilyavolodin:
Here's top 40 browser stats for one week (Feb 12 - 18):

| Sessions | Browser Major Version | Percentage  |
|----------|-----------------------|-------------|
| 40645    | 72 Chrome             | 0.447563151 |
| 29380    | 71 Chrome             | 0.3235184   |
| 4492     | 65 Firefox            | 0.049463739 |
| 3801     | 12 Safari             | 0.04185478  |
| 2352     | 70 Chrome             | 0.02589909  |
| 1063     | 69 Chrome             | 0.011705244 |
| 627      | 68 Chrome             | 0.006904222 |
| 591      | 63 Chrome             | 0.006507807 |
| 550      | 67 Chrome             | 0.006056335 |
| 529      | 74 Chrome             | 0.005825093 |
| 522      | 17 Edge               | 0.005748012 |
| 520      | 73 Chrome             | 0.005725989 |
| 502      | 66 Firefox            | 0.005527782 |
| 491      | 11 Internet Explorer  | 0.005406655 |
| 418      | 58 Opera              | 0.004602815 |
| 381      | 64 Firefox            | 0.004195388 |
| 297      | 57 Opera              | 0.003270421 |
| 285      | 11 Safari             | 0.003138283 |
| 268      | 60 Firefox            | 0.002951087 |
| 260      | 65 Chrome             | 0.002862995 |
| 253      | 66 Chrome             | 0.002785914 |
[...]

PR #588, opened 2019-07-01
Chore: add build step to bundle client-side assets

kaicataldo:
refs https://github.com/eslint/eslint.github.io/issues/576

~I believe I have everything working except for the styling of the JS in the demo. `orion/editor/built-editor.js` relies on RequireJS to request `orion/editor/stylers/application_javascript/syntax`, and it's currently not able to find the module.~ I dug into this some more, and Orion `require`s the syntax highlighting file asynchronously at run time. I struggled with it for a bit and found it easier to just switch it out for CodeMirror (the Babel REPL uses this as well).

There are a lot of ways we can organize the code, and after exploring a few options this felt like the solution with the least amount of developer friction and overhead. It gets rid of some top level directories (I'd ideally like to move a lot of these top level directories into sub-directories to clean up the root directory of the repo) and puts all the client-side assets in one `assets/` directory (which is what  [Jekyll suggests](https://jekyllrb.com/docs/step-by-step/07-assets/)). I didn't rename `styles` to `css`, but we definitely can.

Happy to discuss other ways of doing things, though, if others have strong opinions!

Landing this opens us up a lot of options to us, and I wanted to keep this (mostly) limited to implementing a build step in a sane way.

Finally, some things we definitely need to do if we land this as-is:

1. We'll also need to update the [release script](https://github.com/eslint/eslint/blob/master/Makefile.js#L777-L778) before the next release. That being said, I think it could be nice to explore instead bundling latest Espree/ESLint in this repository at build time in the future.

1. We'll need to change the command Netlify runs before this is merged so that it'll build correctly (note that this is why the deployment preview is broken - neither the styles nor the JS are being built right now).

Note: The design is _slightly_ different for the demo, but I think it's a little cleaner looking:
 - I removed the border because I think it's a little cleaner looking
 - I changed the error underline from a red squiggly line to a red dotted line because the squiggly line wouldn't show up correctly when highlighting one character (which we do for a syntax error)

See updated design below:
Screen Shot 2019-07-05 at 6 52 46 PM


.babelrc
```diff
@@ -0,0 +1,11 @@
+{
+    "presets": [
+      "@babel/preset-react",
+      ["@babel/preset-env", {
+        "modules": false,
+        "targets": {
+          "browsers": ["last 3 versions"]
```
kaicataldo: What do we want to do for this? Relevant discussion: https://github.com/eslint/eslint.github.io/issues/543

.eslintrc.js
```diff
@@ -0,0 +1,77 @@
+"use strict";
+
+module.exports = {
+    root: true,
+    extends: ["eslint"],
+    parserOptions: {
+        ecmaVersion: 2019
```
kaicataldo: I'd ideally like to start using ES2015+ features, but I do think it's important we stick to features available in Espree so that we can dogfood it in our site. Does limiting ourselves to features available in Espree (so, features that have reached stage 4 and are supported by the parser) sound like a good policy?

_config.dev.yml
```diff
@@ -15,6 +15,7 @@ exclude:
   - Gemfile.lock
   - package.json
   - node_modules
+  - src
```
kaicataldo: Since it's not ignoring `assets/build`, this ensures Jekyll waits to regenerate the site until after Webpack finishes and outputs the bundles.

_config.yml
```diff
@@ -2,8 +2,7 @@ title: "ESLint - Pluggable JavaScript linter"
 repository: "eslint/eslint.github.io"
 exclude: [
   'node_modules',
-  'bower_components',
-  'vendor'
+  'src'
```
kaicataldo: Since it's not ignoring assets/build, this ensures Jekyll waits to regenerate the site until after Webpack finishes and outputs the bundles (I don't think this really matters what order they run in, but this prevents Jekyll from regenerating the site when `src` files are edited and then once more when the bundles are outputted).

_tools/fetch-sponsors.js
```diff
@@ -13,7 +13,7 @@
 //-----------------------------------------------------------------------------
 
 const fs = require("fs");
-const fetch = require("node-fetch");
+const fetch = require("node-fetch"); // eslint-disable-line node/no-unpublished-require
```
kaicataldo: I'm not sure why this is warning at the moment. Any ideas?

_tools/fetch-team-data.js
```diff
@@ -13,7 +13,7 @@
 //-----------------------------------------------------------------------------
 
 const fs = require("fs");
-const octokit = require("@octokit/rest")();
+const octokit = require("@octokit/rest")(); // eslint-disable-line node/no-unpublished-require
```
kaicataldo: I'm not sure why this is warning at the moment. Any ideas?

package.json
```diff
@@ -26,19 +33,28 @@
   },
   "homepage": "https://github.com/eslint/eslint.github.io#readme",
   "dependencies": {
-    "moment": "^2.24.0"
+    "moment": "^2.24.0",
```
kaicataldo: I don't think this is actually used anywhere...Will remove.

_tools/fetch-sponsors.js
g-plane: `node-fetch` is declared at `devDependencies`, so the `eslint-plugin-node` reported it.

kaicataldo: That makes sense!

webpack.config.js
```diff
@@ -0,0 +1,28 @@
...
+const path = require("path");
+
+module.exports = {
+    entry: {
+        demo: ["core-js/stable", "regenerator-runtime/runtime", path.join(__dirname, "src/js/demo/index.js")],
+        parser: ["core-js/stable", "regenerator-runtime/runtime", path.join(__dirname, "src/js/parser/index.js")]
+    },
+    output: {
+        path: path.join(__dirname, "assets/build/js"),
+        filename: "[name].js"
```
g-plane: Do we need content hash here to avoid browsers cache issue?

kaicataldo: That's a good question. Going off [this blog post by Netlify](https://www.netlify.com/blog/2017/02/23/better-living-through-caching/), it seems like we shouldn't need to. This seems to be the [same conclusion the Babel team came to](https://github.com/babel/website/blob/master/webpack.config.js#L10-L11).

package.json
ilyavolodin: This was used in the scripts that were pulling sponsor info, I think now that we've switched to GraphQL API, it's no longer needed.

.babelrc
kaicataldo: If we don't wan to think about this yet, we can also specify `forceAllTransforms: true`, and it will compile everything to be ES5-compatible.

.eslintrc.js
kaicataldo: I think we should in the README (and maybe also create a CONTRIBUTING.md document)

webpack.config.js
```diff
@@ -0,0 +1,48 @@
...
+    module: {
+        rules: [
+            {
+                test: /\.jsx?$/,
+                loaders: ["babel-loader"],
+                include: path.join(__dirname, "src/js"),
+                exclude: /node_modules|src\/js\/vendor/
+            },
+            {
+                test: /\.less$/,
```
kaicataldo: Letting Webpack bundle these styles with their relevant bundles seemed like an easier way to do this (especially now that demo also requires the CodeMirror styles). Thoughts?

_config.yml
```diff
@@ -1,9 +1,9 @@
 title: "ESLint - Pluggable JavaScript linter"
 repository: "eslint/eslint.github.io"
-exclude: [
-  'node_modules',
-  'src'
-]
+exclude:
+  - node_modules
+  - src
+  - vendor
```
kaicataldo: This fixes the Travis CI build

_tools/fetch-sponsors.js
kaicataldo: I know this isn't the rule's current behavior, but it seems like it shouldn't report when the root package.json has `private: true`?

kaicataldo:
Putting the do not merge label on this since the deploy preview version of the site isn't able to load all assets at the moment.

ilyavolodin:
@kaicataldo Could you possibly separate CodeMirror into it's own PR that we can merge before we do this one? I would like to see those changes before we do a big-bang build changes.

kaicataldo:
@ilyavolodin I do generally try to decouple tasks like this, but the reason I switched out Orion for CodeMirror is because it was difficult to get Orion to play nicely with webpack and ES modules.

I can spend time getting CodeMirror to work with the RequireJS setup we currently have (maybe by using a CDN?), but I’m not sure it buys us much if we’re going to be adding the bundling step soon anyway (unless you foresee us not merging this PR for a while).

kaicataldo:
As a follow up to  deploy preview version not being able to load assets - it's because the URLs are using `{{ site.url }}` and so the staging environment is requesting assets from `eslint.org`. I don't think we have to fix deploy previews in this PR (and would rather decouple that from this task and land this first), but we could maybe use something like https://github.com/jayvdb/jekyll-netlify to access the deploy preview URL.

_tools/fetch-sponsors.js
kaicataldo: @mysticatea Does that make sense? Maybe I'm missing something.

kaicataldo:
I know this is a large PR, but a lot of it is actually the removal of files and package-lock.json changes. I think I did a pretty good job of isolating individual tasks in commits, and I think reviewing commit by commit makes it a lot easier to review.

_tools/fetch-sponsors.js
aladdin-add: can we disable these rules in eslintrc? (since it is not a node lib.)

mysticatea: If you have `files: [...]` in your package.json, with publish files/dirs, that error should disappear.

kaicataldo: @aladdin-add’s suggestion makes sense to me. Will disable for the website.

kaicataldo:
As discussed via email, I'm going to go ahead and merge this. Happy to continue discussing any of these decisions!

PR #593, opened 2019-07-09
Fix: add Jekyll plugin to fix deployment preview URLs

kaicataldo:
Currently, deployment preview builds are broken because we use `{{ site.url }}` in our Liquid templates to generate URLs. According to the [Netlify docs](https://www.netlify.com/docs/continuous-deployment/#environment-variables), it looks like the deployment preview URL should be set to this environment variable.

Please look over my Ruby - I'm not particularly familiar with the language :)

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -6,7 +6,6 @@ exclude: [
   'vendor'
 ]
 description: "A pluggable and configurable linter tool for identifying and reporting on patterns in JavaScript. Maintain your code quality with ease."
-url: https://eslint.org
 permalink: /blog/:year/:month/:title
 plugins:
     - jekyll-sitemap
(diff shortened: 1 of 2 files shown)


kaicataldo:
@platinumazure Thanks for the quick review! Mind looking it over one more time? I didn't realize Netlify also sets a `URL` env var that should point to the custom domain.

g-plane:
Does it also behave correctly at preview mode?

kaicataldo:
@g-plane I’m not sure I follow. Mind clarifying what you mean by preview mode?

g-plane:
For example, we can preview the website if a pull request is opened. At this case, will the URL be correct?

kaicataldo:
@g-plane Yeah, that’s the purpose of this change! You can test it out yourself using the netlify/eslint/deploy-preview check in this PR.

Issue #590, opened 2019-07-05
TypeError on home page


kaicataldo:
<img width="1358" alt="Screen Shot 2019-07-05 at 6 31 11 PM" src="https://user-images.githubusercontent.com/7041728/60747438-1e3e1d80-9f53-11e9-84ef-60e17a22662d.png">

The following error shows up in the console for the home page only (this script is guarded in the templates so that it only executes on the home page): 

```
Uncaught TypeError: Cannot read property 'appendChild' of null
    at HTMLDocument.<anonymous> ((index):363)
```

Taking a quick look, it appears that there isn't an element with the class name `.logos` on the page when it executes. Relevant code [here](https://github.com/eslint/eslint.github.io/blob/master/_includes/footer.html#L63).

PR #595, opened 2019-07-11
Chore: consolidate and bundle inline scripts

kaicataldo:
fixes https://github.com/eslint/eslint.github.io/issues/590

This PR consolidates the remaining inline scripts we have into a `main.js` bundle. The only remaining inline script we have is in `/docs/user-guide/formatters/html-formatter-example.html`.

This PR does a few things:
1) I've removed some of the broken scripts that no longer do anything (please see comments below).
2) We can now lint these scripts!
4) We were only loading the lazy loading script on the home page, but we actually try to use it in a few of the blog posts. This fixes that.


_includes/footer.html
```diff
@@ -17,100 +15,9 @@
           <li>Copyright JS Foundation and other contributors, <a href="https://js.foundation">https://js.foundation/</a></li>
         </ul>
       </footer>
-    </div><!-- /.container -->
-
-  <link href="https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,600" rel="stylesheet">
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/bootstrap.native/2.0.15/bootstrap-native.min.js" defer></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/anchor-js/4.1.0/anchor.min.js" defer></script>
```
kaicataldo: Now installed via npm and bundled by webpack

_includes/footer.html
```diff
@@ -17,100 +15,9 @@
...
-    <!-- at the end of the BODY -->
-  <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
```
kaicataldo: Now installed via npm and bundled by webpack

_includes/footer.html
```diff
@@ -17,100 +15,9 @@
...
-        for (var i = array.length - 1; i > 0; i--) {
-          var j = Math.floor(Math.random() * (i + 1));
-          var temp = array[i];
-          array[i] = array[j];
-          array[j] = temp;
-        }
-        return array;
-      }
-
-      var logosNode = document.querySelector('.logos');
```
kaicataldo: This is left over from when we had the "Who uses ESLint?" logos on the home page. Since those were removed, this script doesn't actually do anything.

_includes/footer.html
```diff
@@ -17,100 +15,9 @@
...
-            '<div class="slide col-sm-4">' +
-              '<a href="' + logos[i].url + '">' +
-                '<img class="lazyload" data-src="' + logos[i].src + '" src="" alt="' + logos[i].name + '">' +
-              '</a>' +
-            '</div>' +
-          '</div>');
-        logosNode.appendChild(logo);
-      }
-
-      var expenderNode = document.querySelector('.expender');
```
kaicataldo: We don't have a `.expender` element anywhere and this causes the next line to throw as well.

_includes/footer.html
```diff
@@ -17,100 +15,9 @@
...
-    })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
-
-    ga('create', 'UA-60915033-1', 'eslint.org');
-    ga('send', 'pageview');
-
-  </script>
+    </div>
 
-  {% if page.demopage %}
-    <script src="{{ site.url }}/assets/build/js/demo.js"></script>
```
kaicataldo: Moved this to `demo/index.html`

src/js/main/index.js
```diff
@@ -0,0 +1,42 @@
...
+
+document.addEventListener("DOMContentLoaded", function() {
+
+    // Initialize deep anchor links for headings
+    const anchors = new AnchorJS();
+
+    anchors.add(".doc h2, .doc h3, .doc h4, .blog h2, .blog h3, .blog h4");
+
+    // Initialize lazy loading images
+    ImageObserver.createObserver();
```
kaicataldo: Load this on every page since we also attempt to lazy load images in the blog as well.

webpack.config.js
```diff
@@ -6,7 +6,8 @@ const CleanCSSPlugin = require("less-plugin-clean-css");
 module.exports = {
     entry: {
         demo: ["core-js/stable", "regenerator-runtime/runtime", path.join(__dirname, "src/js/demo/index.js")],
-        parser: ["core-js/stable", "regenerator-runtime/runtime", path.join(__dirname, "src/js/parser/index.js")]
+        parser: ["core-js/stable", "regenerator-runtime/runtime", path.join(__dirname, "src/js/parser/index.js")],
+        main: ["core-js/stable", "regenerator-runtime/runtime", path.join(__dirname, "src/js/main/index.js")]
```
kaicataldo: There must be a better way to bundle `core-js/stables` and `regenerator-runtime/runtime` than this. It's a little awkward because the demo/parser scripts are loaded first.

cc @hzoo Any ideas of how we could make this better (ideally a separate bundle that can be cached by the browser?).

kaicataldo: Going to merge this as is for now and explore this in a future PR.

kaicataldo: Related PR: https://github.com/eslint/eslint.github.io/pull/598

PR #603, opened 2019-07-12
Chore: bundle bootstrap.native using webpack

kaicataldo:
refs https://github.com/eslint/eslint.github.io/pull/600, https://github.com/eslint/eslint.github.io/pull/602

Trying this again. Sorry for missing this earlier!

### DIFF
--- a/src/js/main/index.js
+++ b/src/js/main/index.js
@@ -1,6 +1,12 @@
 import "core-js/stable";
 import "regenerator-runtime/runtime";
 
+/*
+ * bootstrap.native automatically initializes components.
+ * Elements that are not on the page when this script loads
+ * will require manual initialization.
+ */
+import "bootstrap.native";
 import docsearch from "docsearch.js";
 import AnchorJS from "anchor-js";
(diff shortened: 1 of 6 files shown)


src/js/main/index.js
```diff
@@ -10,6 +16,10 @@ const htmlClassList = document.documentElement.classList;
 htmlClassList.remove("no-js");
 htmlClassList.add("js");
 
+// The menu dropdown toggles are also links
+document.querySelectorAll("[data-toggle=\"dropdown\"]")
```
kaicataldo: The older version of `bootstrap.native` we were using would handle this for us.

src/js/main/index.js
```diff
@@ -1,6 +1,12 @@
 import "core-js/stable";
 import "regenerator-runtime/runtime";
 
+/*
+ * bootstrap.native automatically initializes components.
+ * Elements that are not on the page when this script loads
+ * will require manual initialization.
+ */
+import "bootstrap.native";
```
kaicataldo: This is what I missed in the last PR.

PR #604, opened 2019-07-12
Fix: set up Netlify redirect rules

kaicataldo:
Redirects are currently broken. This PR sets up redirect rules for Netlify (see docs [here](https://www.netlify.com/docs/redirects/)).

Would love if someone who has some more historical context around these redirects could look these over!

### DIFF
--- a/docs/0.24.1/user-guide/integrations.md
+++ b/docs/0.24.1/user-guide/integrations.md
@@ -1,7 +1,6 @@
 ---
 title: Integrations
 layout: doc
-redirect_from: "/docs/integrations/"
 ---
 
 # Integrations
@@ -30,7 +29,7 @@ redirect_from: "/docs/integrations/"
 * **Broccoli:**
     * [broccoli-eslint](https://www.npmjs.org/package/broccoli-eslint)
 * **Webpack:**
-    * [eslint-loader](https://www.npmjs.org/package/eslint-loader) 
+    * [eslint-loader](https://www.npmjs.org/package/eslint-loader)
 * **Ember-cli**
     * [ember-cli-eslint](https://www.npmjs.com/package/ember-cli-eslint)
 * **Sails.js**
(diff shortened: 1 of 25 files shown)


docs/0.24.1/user-guide/integrations.md
```diff
@@ -1,7 +1,6 @@
 ---
 title: Integrations
 layout: doc
-redirect_from: "/docs/integrations/"
```
kaicataldo: Does anyone know why this is set for so many different versions?

docs/1.0.0/user-guide/integrations.md
```diff
@@ -1,7 +1,6 @@
 ---
 title: Integrations
 layout: doc
-redirect_from: "/docs/integrations/"
```
kaicataldo: Does anyone know why this is set for so many different versions?

docs/1.10.3/user-guide/integrations.md
```diff
@@ -1,7 +1,6 @@
 ---
 title: Integrations
 layout: doc
-redirect_from: "/docs/integrations/"
```
kaicataldo: Does anyone know why this is set for so many different versions?

docs/2.0.0/user-guide/integrations.md
```diff
@@ -1,7 +1,6 @@
 ---
 title: Integrations
 layout: doc
-redirect_from: "/docs/integrations/"
```
kaicataldo: Does anyone know why this is set for so many different versions?

docs/0.24.1/user-guide/integrations.md
ilyavolodin: I think that's because we want to always show latest integration, and not version specific ones. But to be honest, that's just a guess. It's been a very long time since I've added those.

PR #605, opened 2019-07-16
Fix: update files copied into _site

kaicataldo:
This PR does three things:
 - I didn't realize files that start with `_` are not included in the build, so `_redirects` isn't currently being included
 - We're serving a number of files that are copied (for examples, https://eslint.org/webpack.config.js). This should fix this. 
 - Removes the `CNAME` file that is only necessary for GitHub Pages.

### DIFF
--- a/_config.dev.yml
+++ b/_config.dev.yml
@@ -9,13 +9,16 @@ future: true
 
 # Don't generate this stuff
 exclude:
-  - README.md
-  - CNAME
-  - Gemfile
-  - Gemfile.lock
-  - package.json
   - node_modules
   - src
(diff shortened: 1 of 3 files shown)


_config.dev.yml
```diff
@@ -9,13 +9,16 @@ future: true
 
 # Don't generate this stuff
 exclude:
-  - README.md
-  - CNAME
```
aladdin-add: is the "CNAME" need to be served?

kaicataldo: Good question! This is only needed for GitHub Pages. I'll remove the file.

PR #609, opened 2019-07-17
Upgrade: react & react-dom@16.8.6

kaicataldo:
I initially set out to update this to the latest version of React and found a few bugs along the way. Most of the changes here are from updating to use native classes and fixing the resulting linting errors. 

The current demo actually breaks when `ecmaVersion < 2015` and `sourceType = 'module'`:
Screen Shot 2019-07-17 at 1 11 58 AM

Additionally, the formatting of the error message doesn't currently account for error messages that do not contain location information:
Screen Shot 2019-07-17 at 1 11 54 AM


src/js/demo/components/Editor.jsx
```diff
@@ -60,27 +59,32 @@ export default class Editor extends React.Component {
...
-                        line: error.line - 1,
-                        ch: error.column - 1
-                    };
-                    to = {
-                        line: error.line - 1,
-                        ch: error.column
-                    };
+
+                    // line and column are undefined when parsing cannot occur (e.g. misconfiguration)
+                    if (typeof error.line === "number" && typeof error.column === "number") {
```
kaicataldo: This is what fixes the demo from crashing.

src/js/demo/components/Message.jsx
```diff
@@ -1,5 +1,15 @@
 import React from "react";
-import events from "./events";
+import events from "../utils/events";
+
+function formatMessage({ line, column, message }) {
+
+    // line and column are undefined when parsing cannot occur (e.g. misconfiguration)
+    if (typeof line === "number" && typeof column === "number") {
+        return `${line}:${column} - ${message}`;
```
kaicataldo: This is the other bug I found. When there isn't  location information, we can just show the message!

PR #642, opened 2019-11-01
Update: Refresh homepage

nzakas:
I was bored and just took a quick pass over the homepage to freshen it up a bit. It's not a complete redesign, which I'd still like to do, but I feel like it's a nice change.

You can view it live using the Netlify deploy preview.


ilyavolodin:
To be honest, I'm not a huge fan of the new starry-night background. But do like content layout changes under it. But I can live with it.

nzakas:
@kaicataldo I’m not sure what’s up with the package-lock.json file. I just did an npm install after check out, so maybe some new point releases came out for some dependencies?

@ilyavolodin I’m not married to that photo, I just wanted something dark with purple highlights that wasn’t a solid color. Happy to change it if you can find something you like better.

ilyavolodin:
If the goal is to make less solid color on the screen, and add more purple accents, I would just blow up background image logo to cover much larger area of the jumbotron. Something like this:
image

kaicataldo:
Re the lockfile: I don't think it should update the actual versions unless the `package.json` changes. Not sure what's going on. I wonder if it would make sense to just revert the `package-lock.json` changes?

platinumazure:
On mobile, there's a small inconsistency around horizontal alignment of header tags. I'm hoping this screenshot will show what I mean:

Screenshot_20191102-014633

Basically, the new content's headers are left-aligned, but the sponsor headers are center-aligned. Not sure what changes should be made (or if we even need to make changes), but I wanted to share my observations just in case.

nzakas:
@ilyavolidin I was looking to make the hero less bland. Making the logo larger doesn’t accomplish that. I want some texture there.

@platinumazure those are different sections, and the logos don’t look good left-aligned, so I’m not too worried. 

Again, the goal here is not to do a whole page redesign, I just wanted to refresh the above the fold design. We can continue to iterate in other parts.

nzakas:
I reverted `package-lock.json`.

nzakas:
If anyone wants to peruse other images to use for the hero, this is where I looked:
https://unsplash.com/s/photos/purple

I think ultimately this will be a matter of personal preference so if we can just decide on one everyone can live with, we can continue looking for replacement images later.

@platinumazure just FYI, I double-checked and, indeed, we have the same left-align/center-align of headings on the homepage right now, so I think we're fine to leave it that way for now.

platinumazure:
@nzakas Thanks for confirming re: heading alignment. I'm okay with this set of changes at this point.

nzakas:
Okay, I'm going to go ahead and merge this since it seems like people generally like photo. We can definitely change it later if anyone finds something better.

PR #645, opened 2019-11-06
New: November fundraising blog post update

nzakas:
This is a blog post update about fundraising. I'd like to get this out tomorrow (07 November), so please take a look!


ilyavolodin:
LGTM, thanks! I think we should also cross-post those updates in OpenCollective.

kaicataldo:
Might be nice to tweet it too!

nzakas:
I'll do both! Thanks everyone.

PR #648, opened 2019-11-14
Update: Add sponsor count and donation totals to homepage

nzakas:
This adds a count of our total contributors and monthly donations to the homepage. I think this is useful because Open Collective, for some reason, doesn't have this info easily available (I end up calculating it by hand every time). 
Screen Shot 2019-11-14 at 10 24 22

### DIFF
--- a/_pages/index.html
+++ b/_pages/index.html
@@ -21,11 +21,23 @@ homepage: true
 </div>
 
 <hr class="hr-near-below">
+{% assign tiers = "gold,silver,bronze,backers" | split: "," %}
+{% assign monthlyTotal = 0 %}
+{% assign supportersCount = 0 %}
+{% for tier in tiers %}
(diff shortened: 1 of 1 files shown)


ilyavolodin:
Quick question. Could this be pre-calculated by the job and provided in the sponsors.json file instead? I think it would be much cleaner then doing it in the template.

nzakas:
It could be, but then I'd need to change the format of the sponsors file. Right now, we assume that every top-level property represents a tier. If I were to add another key, then we'd need to change the logic that processes the data both for the website and the ESLint readme. I'm not sure it's worth that much code churn just to output these numbers.

ilyavolodin:
Ok, sounds good. We might need to change format in the future once we enable collecting data from Github, so we could do that at that time.

nzakas:
Sounds good.

PR #649, opened 2019-11-18
Chore: switch from Jekyll to Eleventy

kaicataldo:
refs https://github.com/eslint/website/issues/536

~~This is a work in progress.~~ I know we haven't officially decided to use Eleventy, but I wanted to give it a shot to see what cutting over would look like. ~~It's not all working yet, but it's looking promising!~~

There might be a few small things I need to fix still, but this is looking good to me. Would love to see what others thinks.

Note:

Eleventy is pretty flexible and allows you to pick the templating engine you want to use. The default is `liquid` (which is what Jekyll uses), which made converting fairly straightforward. If we go this route, going forward we'll have more flexibility in terms of templating engines and overall project structure.


ilyavolodin:
Just to make sure that I understand it correctly. The biggest advantage of switching to this is that we don't have go through the pain of installing Jekyll any longer, right? Cause otherwise, it looks pretty much the same.

kaicataldo:
Yes, biggest benefit is that we'll be using a build tool that we can configure using JS. We'll consequently get rid of our dependency on Ruby. Build times also seem to be improved for me.

kaicataldo:
I'll also add that this will give us a lot more flexibility in the future. It looks a lot like Jekyll right now for the sake of conversion.

ilyavolodin:
I see. Well getting rid of Jekyll is a very powerful reason, I absolutely hate it:-)

kaicataldo:
This is mostly working, if anyone else wants to play around with it! Known issues:

- [ ] The fixable link at the top of every rule is broken. It's unfortunately generated in ESLint in the site generation script. I'll add a little hack on this end but we should change that if we decide to land this.

kaicataldo:
I'm not sure why Netlify isn't running...

kaicataldo:
Ah, looks like it's running now :+1: Should be able to check the deploy preview to test it.

mysticatea:
:+1:

kaicataldo:
Some other notes:

1. Eleventy doesn’t allow YAML for data by default. There seem to be workarounds, but they involve monkeypatching and I’d like to avoid that.
1. It might be fixable with some more digging, but I had trouble rendering HTML in liquid loops in Markdown. It would output the the correct markup, but it was HTML encoded and so would render as a string. I ended up rewriting the problematic Markdown files as Liquid files, and I actually think it’s easier to read.

kaicataldo:
Okay, I think I have everything working, down to syntax highlighting (`highlight.js`'s `github` theme is pretty much identical to the one we're currently using - I saw almost because some characters aren't bolded, but otherwise it seems to be the same). 🎉 

Ready for evaluation from the team! Please let me know if you see anything amiss.

I turned some of the YAML data files into JSON files because Eleventy doesn't support YAML for data out of the box. I would want to make sure I update the data generation script in eslint/eslint (at least until we try [moving site generation into this repo](https://github.com/eslint/website/issues/596).

platinumazure:
I haven't the time to review the PR itself, but I looked at the Netlify preview and everything looks nice!

Things I looked at:

* Home
* Rules list
* Rule details
* Formatters list/"demos"
* Node.js API
* Blog lists
* Blog post
* Espree demo
* ESLint demo
    * Code edits, configuration edits, autofix, URL persistence

kaicataldo:
Thanks for taking a look! I just noticed some of the related rules links are broken on the rules pages.

mysticatea:
Looks like the cause is trailing slash.

`/docs/rules/no-labels` is redirecting to `/docs/rules/no-labels/`, so the relative paths in the page point wrong hierarchy. I think that we should do the reverse of the redirecting because `/docs/rules/no-labels` is not a directory.

kaicataldo:
You're right - it's caused by the trailing slash that the deploy preview server is adding to the URL. I wonder if this is just a difference between deploy previews and production?

kaicataldo:
I've confirmed this works fine in our current setup with Jekyll in deploy previews, so it's something we'll need to fix here.

kaicataldo:
Per [their docs](https://docs.netlify.com/routing/redirects/redirect-options/#trailing-slash), Netlify doesn't allow you to use redirects to remove slashes. However, it looks like we can fix this by generating the named HTML files directly instead of nesting them in a directory, because we're using Netlify's "Pretty URLs" feature.

kaicataldo:
This has been fixed. The only downside of this is that the default server that's built into Eleventy adds trailing slashes, so dev is broken. I'll add a small Express server we can use in dev and we should be all sorted then.

kaicataldo:
Dev should be fixed.

kaicataldo:
I'm having difficulty getting blog post links to work locally, even though they work in the deploy preview environment. Maybe someone else has an idea?

Currently, `http://localhost:8080/blog/2019/10/eslint-v6.6.0-released` is returning a `404`. However, `http://localhost:8080/blog/2019/10/eslint-v6.6.0-released.html` resolves to the correct file. Any idea why this would be? Both locally and in the deploy preview environment, `http://localhost:8080/blog/` always has a trailing slash, so that doesn't seem to be the issue. 

The one clue I have is that this only occurs with `.` characters - so maybe that has something to do with it?

kaicataldo:
It's definitely the `.` character in the file name. Changing `_posts/2019-10-25-eslint-v6.6.0-released.md` to `_posts/2019-10-25-eslint-v6-6-0-released.md` fixes it.

kaicataldo:
I worked around the above issue (which seems like an upstream bug in Express) by writing a custom route handler that mimics the behavior of 

```js
express.static(path.join(__dirname, "../_site"), { extensions: ["html"] }))
```
with a TODO to try to figure it out later.

I believe this is finally in a mergeable state and would love feedback.

kaicataldo:
Adding "do not merge" because, while this should be ready to go, I'll need to update the scripts that generate data to output JSON instead of YAML before this will work.

platinumazure:
@kaicataldo Any suggestions on how to review this efficiently? Do we need to look at the full diff, or can we ignore some of the files? Are there any files/diffs that you view as key to this change?

kaicataldo:
Good question! I think the most important feedback I’m looking for is if others support this change. Feedback around if this is an improved developer experience would also be great, if others have some time to pull down this branch and try out the local dev setup. Beyond that, I would appreciate eyes on any JS I wrote.

kaicataldo:
I'm also happy to just merge this and own any issues that crop up, if others don't have strong feelings either way. The one external thing that needs to be done before merging this is to update the YAML data files that are generated on release, as this is currently expecting JSON.

nzakas:
It might be easier to create a script that converts YAML data files into JSON before the website is built. That way, we can decouple making this change from changes in the main repo. 

At that point, I’d say make sure the instructions to run in dev work correctly and feel free to merge.

kaicataldo:
Ah, I actually might have been wrong about it not supporting YAML. Had trouble finding info in the docs, but I came across [this](https://github.com/11ty/eleventy/commit/2c7bde7bad05b37c07f65ae3b14c20abdecfd7a2). Will try it out when I get back!

kaicataldo:
I guess that's coming in the next release. I've added a script to convert YAML -> JSON in the interim.

kaicataldo:
I believe this is ready to go! Will deploy sometime in the next few days.

_pages/feed.liquid
```diff
@@ -15,8 +15,8 @@ permalink: feed.xml
         <description>{{ post.templateContent | xmlEscape }}</description>
         {% comment %} https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#valid-date-string {% endcomment %}
         <pubDate>{{ post.date | dateFormat: "yyyy-LL-dd" }}</pubDate>
-        <link>{{ site.url }}{{ post.url }}</link>
-        <guid isPermaLink="true">{{ site.url }}{{ post.url }}</guid>
+        <link>{{ site.url }}{{ post.url | replace: '.html', '' }}</link>
```
mysticatea: Do we want to remove `.html` from the `<link rel="canonical">`, `<meta property="og:url">`, and `<meta name="twitter:url">` of every page as well?

kaicataldo: This should be fixed now!

ilyavolodin:
@kaicataldo I pulled down this branch to try running it locally, but I can't seem to be able to either build or start anything. Here's the error below:
```
Converting YAML files found with glob "'_data/*.yml'" to JSON.
(node:20908) ExperimentalWarning: The fs.promises API is experimental
Trace: No files found for glob '_data/*.yml'.
    at c:\Users\Ilya\Documents\GitHub\eslint.github.io\_tools\yaml-to-json.js:37:17
```
I checked, `_data` directory does contain yml files, so I'm not sure why it's complaining.

kaicataldo:
@ilyavolodin Thanks for checking this out! What Node version are you using?

ilyavolodin:
I'm on 10.14.1 currently. I was able to manually compile everything and run it (bypassing npm scripts, which are still failing). Everything looks fine to me. But just to make sure, I'm currently running the site through crawler, to catch errors.

kaicataldo:
Hmmm I'm unable to replicate locally on 10.17.0 on macOS. I wonder if it's a cross-compatibility issue in early versions of Node? I'll see if I can replicate on my Windows machine.

kaicataldo:
Would be good to start running CI in a Windows env in this repo as well.

kaicataldo:
I was able to replicate this in Windows 10 using Node 12.13.1. Probably some cross-compatibility issue.

package.json
```diff
@@ -1,64 +1,70 @@
...
+    "build:webpack": "webpack --mode=production",
+    "build:less": "lessc src/styles/main.less _site/assets/styles/main.css --clean-css --source-map=_site/assets/styles/main.css.map",
+    "build": "npm run build:eleventy && npm run build:less && npm run build:webpack",
+    "start:eleventy": "npm run convert-yaml && NODE_ENV=development eleventy --watch",
+    "start:webpack": "webpack -w --mode=development",
+    "start": "node _tools/dev-server.js",
+    "lint": "eslint --ext=.js,.jsx .",
+    "test": "npm run lint",
+    "fix": "npm run lint -- --fix",
+    "convert-yaml": "node _tools/yaml-to-json.js '_data/*.yml'"
```
mysticatea:
```suggestion
    "convert-yaml": "node _tools/yaml-to-json.js \"_data/*.yml\""
```

In cmd.exe, single quotes are not a special character to enclose strings.

kaicataldo: I was just debugging this and noticed the same thing! Will fix.

package.json
```diff
@@ -1,64 +1,70 @@
...
   "homepage": "https://github.com/eslint/eslint.github.io#readme",
+  "engines": {
+    "node": ">=12.0.0"
+  },
+  "scripts": {
+    "build:eleventy": "npm run convert-yaml && eleventy",
+    "build:webpack": "webpack --mode=production",
+    "build:less": "lessc src/styles/main.less _site/assets/styles/main.css --clean-css --source-map=_site/assets/styles/main.css.map",
+    "build": "npm run build:eleventy && npm run build:less && npm run build:webpack",
+    "start:eleventy": "npm run convert-yaml && NODE_ENV=development eleventy --watch",
```
mysticatea: In cmd.exe, it doesn't support `VAR=value command` style notation. Maybe [cross-env](https://www.npmjs.com/package/cross-env) solves it.

ilyavolodin:
Crawler finished testing local site. Everything looks pretty good. A few broken links, but I'm pretty sure those are broken in production as well. Here's the output:
```
http://localhost:8080/docs/maintainer-guide/governance links to missing page. List of links is ["http://localhost:8080/docs/maintainer-guide/contributing"]
Error 404 for http://localhost:8080/docs/maintainer-guide/contributing
http://localhost:8080/docs/6.0.0/user-guide/configuring links to missing page. List of links is ["http://localhost:8080/docs/6.0.0/rules/"]
Error 404 for http://localhost:8080/docs/6.0.0/rules/
http://localhost:8080/docs/5.0.0/developer-guide/working-with-rules links to missing page. List of links is ["http://localhost:8080/docs/5.0.0/rules/","http://localhost:8080/docs/5.0.0/rules/"]
Error 404 for http://localhost:8080/docs/5.0.0/rules/
http://localhost:8080/docs/5.0.0/user-guide/configuring links to missing page. List of links is ["http://localhost:8080/docs/5.0.0/rules/"]
http://localhost:8080/docs/4.0.0/user-guide/command-line-interface links to missing page. List of links is ["http://localhost:8080/docs/4.0.0/user-guide/formatters/","http://localhost:8080/docs/4.0.0/user-guide/formatters/","http://localhost:8080/docs/4.0.0/user-guide/formatters/","http://localhost:8080/docs/4.0.0/user-guide/formatters/","http://localhost:8080/docs/4.0.0/user-guide/formatters/","http://localhost:8080/docs/4.0.0/user-guide/formatters/","http://localhost:8080/docs/4.0.0/user-guide/formatters/","http://localhost:8080/docs/4.0.0/user-guide/formatters/","http://localhost:8080/docs/4.0.0/user-guide/formatters/","http://localhost:8080/docs/4.0.0/user-guide/formatters/","http://localhost:8080/docs/4.0.0/user-guide/formatters/","http://localhost:8080/docs/4.0.0/user-guide/formatters/"]
Error 404 for http://localhost:8080/docs/4.0.0/user-guide/formatters/
http://localhost:8080/blog/2016/02/eslint-v2.0.0-released links to missing page. List of links is ["http://localhost:8080/docs/developer-guide/code-path-analysis.html"]
Error 404 for http://localhost:8080/docs/developer-guide/code-path-analysis.html
http://localhost:8080/docs/2.0.0/developer-guide/working-with-rules links to missing page. List of links is ["http://localhost:8080/docs/2.0.0/developer-guide/code-path-analysis"]
http://localhost:8080/docs/2.0.0/developer-guide/code-path-analysis links to missing page. List of links is ["http://localhost:8080/docs/2.0.0/developer-guide/code-path-analysis/"]
Error 404 for http://localhost:8080/docs/2.0.0/developer-guide/code-path-analysis/
http://localhost:8080/blog/2015/06/eslint-0.24.0-released links to missing page. List of links is ["http://localhost:8080/blog/2015/preparing-for-1.0.0"]
Error 404 for http://localhost:8080/blog/2015/preparing-for-1.0.0
Total pages: 939. Errors found 7
```

kaicataldo:
Okay, this is now working for me in Windows 10 on Node 12.13.1. @ilyavolodin Can you check if it all looks good on your end?

ilyavolodin:
`npm run build:eleventy` now works without errors. However, `npm run start` and `npm run start:eleventy` are still erroring out with the same error as before.

kaicataldo:
Could this be an issue with Node 10? I actually set the `engines` as `>=12.0.0`, since that's latest LTS and I figured this is just for local dev. If that doesn't seem like a good idea, I can change engines to `>=10.0.0` and see if I can debug this issue (might not be able to use `fs.promises`).

ilyavolodin:
Everything is working fine for me now. Sorry about the noise, I forgot to run `npm install` after I pulled latest changes. `npm run start` now correctly starts local environment. It's still not blazingly fast, but it is much faster then Jekyll (not a huge achievement, turtles are faster then Jekyll:-)

kaicataldo:
No problem! Appreciate the help :)

kaicataldo:
Yes, I'd love to make it faster in the future. Note that in Jekyll we were excluding a bunch of files from the build to speed it up, which doesn't seem possible with Eleventy at the moment. Would love to figure that out going forward. I'm also hoping Eleventy will eventually get incremental build support - right now it rebuilds the whole world on every file change.

PR #663, opened 2019-12-02
Docs: Fix a broken anchor link

revelt:
Docs: Fix a broken anchor link


kaicataldo:
Thanks for the PR! This file actually gets copied over here from https://github.com/eslint/eslint/blob/master/docs/developer-guide/working-with-rules.md, and so changes need to happen in that file. Would you like to make a PR in https://github.com/eslint/eslint?

PR #665, opened 2019-12-05
Chore: bundle ESLint during build step

kaicataldo:
Here's a PoC for https://github.com/eslint/website/issues/596. Thoughts on doing it this way? Note that I based it off of https://github.com/eslint/website/pull/664 so that we could see what the final configuration would look like. Please ignore the first commit!


src/js/demo/components/App.jsx
```diff
@@ -6,7 +6,7 @@ import BugReport from "./BugReport";
 import Crash from "./Crash";
 import Configuration from "./Configuration";
 import Unicode from "../utils/unicode";
-import linterModule from "../../eslint";
+import linterModule from "../node_modules/eslint/lib/linter/linter";
```
kaicataldo: I thought it was better to be explicit here because we don't ever want to accidentally install the top level eslint package we're using for linting.

webpack.config.js
```diff
@@ -5,20 +5,26 @@ const CleanCSSPlugin = require("less-plugin-clean-css");
...
             {
                 test: /\.jsx?$/,
-                loaders: ["babel-loader"],
-                include: path.join(__dirname, "src/js"),
-                exclude: /node_modules|src\/js\/vendor/
+                loader: "babel-loader",
+                include: path.resolve(__dirname, "src/js"),
+                exclude: [
+                    path.resolve(__dirname, "node_modules"),
+                    path.resolve(__dirname, "src/js/demo/node_modules/lodash"),
```
kaicataldo: From https://github.com/eslint/eslint/blob/master/webpack.config.js#L23

webpack.config.js
```diff
@@ -5,21 +5,26 @@ const CleanCSSPlugin = require("less-plugin-clean-css");
...
             {
                 test: /\.jsx?$/,
-                loaders: ["babel-loader"],
-                include: path.join(__dirname, "src/js"),
-                exclude: /node_modules|src\/js\/vendor/
+                loader: "babel-loader",
+                include: path.resolve(__dirname, "src/js"),
+                exclude: [
+                    path.resolve(__dirname, "node_modules"),
+                    path.resolve(__dirname, "src/js/demo/node_modules/lodash")
```
kaicataldo: From https://github.com/eslint/eslint/blob/master/webpack.config.js#L23

ilyavolodin: I think we need to make sure that webpack is going to be doing tree-shaking if we are bundling lodash. Lodash is huge, and if we pull the whole thing, that's going to blow up the size of the bundle by a very large percentage. We need to make sure that only the modules that we are actually using are bundled (just ran into the same exact issue at work, so want to make sure that we verify this before merging).

kaicataldo: This is only excluding it from running `babel-loader` on the `lodash` files. I do think it's worth taking a look to see if we're including all of `lodash` in our client-side assets here, but this should be the same as it is currently.

PR #666, opened 2019-12-10
Update: add Platinum Sponsorship Tier

kaicataldo:
I haven't yet tested the script locally, however I have manually tested the template logic, and it works with the current data set (no `platinum` key), an empty `platinum` property, and a `platinum` property populated with logos.


kaicataldo:
I have successfully run this version of the script locally. I think we're good to go. Would you prefer for me to commit the results of running it to double check?

ilyavolodin:
No, I think verifying that it runs correctly is perfectly fine.

PR #672, opened 2019-12-16
Chore: add analytics to sponsor links on home page

kaicataldo:
Small PR adding some Google Analytics events to the sponsor links on our home page. I'm not seeing the events fire locally (I'm actually not seeing any analytics requests going out), and I'm assuming that's due to it the requests not originating at `eslint.org`. Would love confirmation though!


ilyavolodin:
Yes, I think that's because of the third attribute when initializing ga points to eslint.org domain. I think we can merge it, and I can check it in GA console, we can always change it if it doesn't work correctly.

kaicataldo:
There was a failure yesterday too that was fixed by a re-run. Will run the deploy-preview deploy once more.

PR #675, opened 2019-12-29
Fix link to section

DamienCassou:


kaicataldo:
Thanks for the PR! Do you mind making this PR in the eslint/eslint repository? This file actually gets copied from there when we run a release. Corresponding file is [here](https://github.com/eslint/eslint/blob/master/docs/developer-guide/working-with-rules.md#contextreport).

DamienCassou:
Done: https://github.com/eslint/eslint/pull/12732

PR #680, opened 2020-01-08
Upgrade: update dependencies 

kaicataldo:
Eleventy v0.10 adds an API that allows us to add YAML files as a data source! This release is also supposed to improve build speed.


.eleventy.js
```diff
@@ -1,12 +1,16 @@
...
 const { addPlugins, addFilters, addTransforms } = require("./_11ty");
+const yaml = require("js-yaml");
 
 module.exports = function(eleventyConfig) {
     addPlugins(eleventyConfig);
     addFilters(eleventyConfig);
     addTransforms(eleventyConfig);
 
+    // Add YAML files as data sources
+    eleventyConfig.addDataExtension("yml", contents => yaml.safeLoad(contents));
```
aladdin-add:
```suggestion
    eleventyConfig.addDataExtension("yml", yaml.safeLoad);
```

kaicataldo: I think we should leave it as-is for now since this is more explicit and a bit safer against either of these APIs changing.

FWIW, the examples are all written this way: https://www.11ty.dev/docs/data-custom/

PR #681, opened 2020-01-16
Update: open demo rule links in new tab

kaicataldo:
I noticed that the links in the demo errors open in the same tab, navigating away from eslint.org. This PR adds a `target="_blank"` attribute to those links to open them in a new tab.

Screen Shot 2020-01-15 at 10 45 59 PM

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -48,6 +48,7 @@ module.exports = {
                 }
             },
             rules: {
+                "react/jsx-no-target-blank": "error",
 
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -48,6 +48,7 @@ module.exports = {
                 }
             },
             rules: {
+                "react/jsx-no-target-blank": "error",
```
kaicataldo: I was surprised this wasn't already on. I had mistakenly thought that this was included with the recommended preset.

PR #702, opened 2020-03-05
chore: fixed small grammar mistakes

anikethsaha:
Made small grammar fixes in `developer-guide/*` pages


nzakas:
Thanks for doing this. Unfortunately, this only fixes the live site. The actual source of these files is in the main ESLint repo. Can you also submit these changes there? (Otherwise, these files will get overwritten in the next release)

anikethsaha:
> Thanks for doing this. Unfortunately, this only fixes the live site. The actual source of these files is in the main ESLint repo. Can you also submit these changes there? (Otherwise, these files will get overwritten in the next release)

sure 👍

PR #713, opened 2020-04-20
Fix: parser interface link

shahroq:


docs/user-guide/configuring.md
```diff
@@ -63,7 +63,7 @@ Setting parser options helps ESLint determine what is a parsing error. All langu
 By default, ESLint uses [Espree](https://github.com/eslint/espree) as its parser. You can optionally specify that a different parser should be used in your configuration file so long as the parser meets the following requirements:
 
 1. It must be a Node module loadable from the config file where it appears. Usually, this means you should install the parser package separately using npm.
-1. It must conform to the [parser interface](/docs/developer-guide/working-with-plugins#working-with-custom-parsers).
+1. It must conform to the [parser interface](/docs/developer-guide/working-with-plugins#processors-in-plugins).
```
kaicataldo, line 66: Thanks for catching this! Agreed that the link is incorrect here. Could we actually link to `/docs/developer-guide/working-with-custom-parsers` instead?

kaicataldo:
Ah, and this change actually needs to happen in the [eslint/eslint](https://github.com/eslint/eslint/blob/master/docs/user-guide/configuring.md) repo, since we copy those files over into this repo when we release the site. Otherwise, this will be overwritten by what's in the other repo.

btmills:
This is now fixed and live! The fix in https://github.com/eslint/eslint/pull/12965 was merged back in February, but it went into the v7 prereleases, so it wasn't live on the site until we shipped v7.0.0 final a few weeks ago. Sorry about the delay there, and thanks for helping to improve ESLint!

PR #719, opened 2020-04-27
chore: Correct examples for no-param-reassign

wdoug:
Specifically for parameter "ignorePropertyModificationsForRegex"

### DIFF
--- a/docs/rules/no-param-reassign.md
+++ b/docs/rules/no-param-reassign.md
@@ -137,15 +137,15 @@ Examples of **correct** code for the `{ "props": true }` option with `"ignorePro
 ```js
 /*eslint no-param-reassign: ["error", { "props": true, "ignorePropertyModificationsForRegex": ["^bar"] }]*/
 
-function foo(bar) {
+function foo(barVar) {
     barVar.prop = "value";
 }
 
-function foo(bar) {
+function foo(barrito) {
     delete barrito.aaa;
 }
 
-function foo(bar) {
+function foo(bar_) {
     bar_.aaa++;
 }
 ```


docs/rules/no-param-reassign.md
```diff
@@ -137,15 +137,15 @@ Examples of **correct** code for the `{ "props": true }` option with `"ignorePro
 ```js
 /*eslint no-param-reassign: ["error", { "props": true, "ignorePropertyModificationsForRegex": ["^bar"] }]*/
 
-function foo(bar) {
+function foo(barVar) {
```
kaicataldo, line 140: Thanks for doing this. Do you mind making this PR over in the https://github.com/eslint/eslint repo? We copy all the documentation from there, so this will get overwritten during our next release.

wdoug: Hmm. Is it [this file](https://github.com/eslint/eslint/blob/master/docs/rules/no-param-reassign.md)? According to that file, this docs issue was already fixed back [in December](https://github.com/eslint/eslint/pull/12492). Has it just not gone through a release yet?

wdoug: If everything gets copied from the https://github.com/eslint/eslint repo, it would be nice to the "Edit this page" links at the bottom of the page point to the other repo instead of this one. I'm trying to wrap my head around the process here.

nzakas: @wdoug it looks like “Edit this page” does go to the correct file. You’ll note there’s an edit_link field at the top of this file indicating the URL to use.

I’m not sure why the change was fixed in the main repo and didn’t end up in the live site. 🤔

kaicataldo: It looks like it was the very first PR to land in `master` when we started the v7.0.0 prereleases (you can see it [here](https://github.com/eslint/eslint/commits/master?after=356fdb46aa118ba3d81cee93e8c058a7c98acaf9+174)), and so this is currently published in the [v7.0.0 docs](https://eslint.org/docs/7.0.0/rules/no-param-reassign). This should be fixed by our release this Friday.

I don't think this is a bug, but an unfortunate side effect of our release process taking so long this major release cycle.

PR #720, opened 2020-04-27
Fix: use forced redirects in _redirects file

kaicataldo:
This PR fixes the redirects that were broken by [the bug fix](https://community.netlify.com/t/changed-behavior-in-redirects/10084) that Netlify implemented. I had been planning to do it before the deadline but then got sick - apologies that it's a bit late. I think marking all of these as forced redirects makes sense since we never want it to follow the shadowing behavior for the existing redirects.

### DIFF
--- a/public/_redirects
+++ b/public/_redirects
@@ -6,7 +6,7 @@ https://eslint.netlify.com/* https://eslint.org/:splat 301!
 
 /docs                               / 301!
 /docs/user-guide/rules              /docs/rules 301!
-/cla                                https://cla.js.foundation/eslint/eslint 301!
+/cla                                https://cla.js.foundation/eslint/eslint 302!
 
 /docs/0.24.1/command-line-interface /docs/user-guide/command-line-interface 301!
 /docs/0.24.1/configuring            /docs/user-guide/configuring 301!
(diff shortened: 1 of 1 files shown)


public/_redirects
```diff
@@ -4,33 +4,33 @@
...
-
-/docs/2.13.1/command-line-interface /docs/user-guide/command-line-interface
-/docs/2.13.1/configuring            /docs/user-guide/configuring
-/docs/2.13.1/integrations           /docs/integrations
-/docs/2.13.1/user-guide/rules       /docs/rules
-
-/docs/3.0.0/user-guide/rules        /docs/rules
+/docs                               / 301!
+/docs/user-guide/rules              /docs/rules 301!
+/cla                                https://cla.js.foundation/eslint/eslint 301!
```
nzakas, line 9: This doesn’t seem like the right URL anymore?

nzakas: And maybe should be a 302 so we can change this whenever we need to?

kaicataldo: A `302` makes sense to me!

Do you know what the URL should be? I see that the link in the checks below points to `https://cla.js.foundation/eslint/website?pullRequest={PR_number}`. Surprised that it's not hosted at `openjsf.org`.

nzakas: I have no idea, a quick search didn’t reveal anything. I suppose we can leave it as-is for now, just with a 302

kaicataldo: Sounds good. I'll follow up with the OpenJSF to confirm what we should be using.

PR #726, opened 2020-05-08
Docs: Add v7 migration guide to header menu

btmills:
Now that we have 7 migration guides in the menu, what do y'all think of putting them in descending order so the most recent version's guide will be listed first?


kaicataldo:
Descending order sounds good to me :+1:

PR #728, opened 2020-05-11
Docs: Change Markdown code fence to <code> tag

btmills:
The current version is rendering the raw backticks.


kaicataldo:
Thanks for fixing this!

PR #733, opened 2020-05-27
Update logos.yml

ndonovic:
Added Toptal


ndonovic:
Note to whoever sees this - all should be in order - I see the commit-message and I added the 2nd commit after that check was run, so all should be in order and properly done at this stage.

nzakas:
We only check the first commit, so you’d need to squash and re-push to fix the failure, but I can do it while merging.

Just a note that your logo will show up on this page in a few minutes: https://eslint.org/users

nzakas:
Oops, looks like I’m unable to merge from my phone due to the failing check. We will get back to it soon.

ndonovic:
Great, thank you @nzakas . Looking forward to it.

nzakas:
Thanks @kaicataldo

PR #735, opened 2020-06-06
Update: Add banner

kaicataldo:
This PR implements a reusable banner we can use when we want to announce something.

Screenshot:
Screen Shot 2020-06-05 at 9 41 43 PM

### DIFF
--- a/_data/site.js
+++ b/_data/site.js
@@ -17,6 +17,11 @@ function calculateUrl() {
 module.exports = {
     title: "ESLint - Pluggable JavaScript linter",
     description: "A pluggable and configurable linter tool for identifying and reporting on patterns in JavaScript. Maintain your code quality with ease.",
+    banner: {
+        text: null,
+        foregroundColor: "#ffffff",
+        backgroundColor: "#000000"
+    },
     url: calculateUrl(),
     links: {
         chat: "/chat",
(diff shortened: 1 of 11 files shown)


src/styles/lib/banner.less
```diff
@@ -0,0 +1,9 @@
+.eslint-banner {
+  font-size: 24px;
+  font-weight: bold;
+  text-align: center;
+  margin: 0 auto;
+  background-color: #000000;
```
kaicataldo, line 6: Sane defaults

_data/site.js
```diff
@@ -17,6 +17,11 @@ function calculateUrl() {
 module.exports = {
     title: "ESLint - Pluggable JavaScript linter",
     description: "A pluggable and configurable linter tool for identifying and reporting on patterns in JavaScript. Maintain your code quality with ease.",
+    banner: {
+        text: null,
```
kaicataldo, line 21: This won't render anything at the moment. I can update this with copy either in this PR or in a follow up once we finalize it.

nzakas: Nice.

btmills: Disclaimer: my understanding of the template syntax is limited. Would we be able to include links within this string? That’s the only improvement I could see wanting if that doesn’t work currently.

kaicataldo: I wasn’t sure either, but I verified this does work!

PR #761, opened 2020-08-03
Fix: Team data fetch (fixes #757)

### LINKED ISSUE #757: ESLint main Readme.md is not updated with new team members
Both @mdjermanovic and @anikethsaha are missing from README.md in the main github repo. Looks like automation failed.

nzakas:
It turns out `@octokit/rest` changed its API completely without a major version bump, so things broke. I upgraded to the latest.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@babel/preset-env": "^7.10.2",
     "@babel/preset-react": "^7.10.1",
     "@octokit/graphql": "^4.5.0",
-    "@octokit/rest": "^17.11.0",
+    "@octokit/rest": "^18.0.3",
     "babel-loader": "^8.1.0",
     "chokidar": "^3.4.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -45,7 +45,7 @@
     "@babel/preset-env": "^7.10.2",
     "@babel/preset-react": "^7.10.1",
     "@octokit/graphql": "^4.5.0",
-    "@octokit/rest": "^17.11.0",
+    "@octokit/rest": "^18.0.3",
```
platinumazure, line 48: Do we want to pin to a minor version (~18.0.3), in case of future breaks?

nzakas: I'd rather leave as-is so we handle the versions through the CLI instead of handcoding just one. This breaking change was because they moved to a more ESM-friendly format, so I don't think we'll go through it again.

PR #766, opened 2020-08-11
New: Add Ethical Ads

nzakas:
This ads Ethical Ads to each page (also removes the banner). The fallback if there is no ad, or if JS is disabled, is to show an ad for sponsoring ESLint.

I have an open question to them about how the ad looks on the homepage as it seems there's some weirdness with the CSS not being caused by us.

### DIFF
--- a/_data/site.js
+++ b/_data/site.js
@@ -18,7 +18,7 @@ module.exports = {
     title: "ESLint - Pluggable JavaScript linter",
     description: "A pluggable and configurable linter tool for identifying and reporting on patterns in JavaScript. Maintain your code quality with ease.",
     banner: {
-        text: "<a href='https://code2040-2020.funraise.org/' style='color: #a486df'>Donate to Code2040</a>",
+        text: null,
         foregroundColor: "#ffffff",
         backgroundColor: "#150b29"
     },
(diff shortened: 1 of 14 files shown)


kaicataldo:
I know this is a WIP - so this might be in the pipeline - but the layout on a few other pages looks a little wonky with the floated ad container as well:

Team page:
Screen Shot 2020-08-12 at 11 46 31 PM

Sponsors page:
Screen Shot 2020-08-12 at 11 45 30 PM

nzakas:
@btmills yeah, that's the homepage weirdness I was talking about. I figured out how to fix it. For the others, if you are only seeing the sponsorship fallback, it's possible you have an ad blocker on your browser. The sponsorship fallback will only be shown if the ad doesn't load for some reason.

@kaicataldo oops, missed those pages. I'll take a look.

nzakas:
Okay, this should be set now. Thanks for the feedback.

kaicataldo:
LGTM! I don't there's enough contrast with the ad on the hero of the home page, but that's an existing problem with the current header text as well. Maybe we can find an alternate background image (or scrap the image altogether), but that shouldn't block this.

nzakas:
Okay, I’ll plan on merging on Monday so we can finalize the setup. No sense in having three days of placeholder ads on the site over the weekend.

PR #768, opened 2020-08-14
Chore: remove extra post file

kaicataldo:
Looks like we missed deleting this file when it was renamed. No harm done - this isn't in the list of blog posts.


nzakas:
I had deleted it and then accidentally introduced it again. Thanks for catching this.

PR #786, opened 2020-09-11
Fix: use parseForESLint API (fixes #785)

### LINKED ISSUE #785: Example for linter.defineParser uses old "parse" API
The code example in the documentation for eslint.defineParser at
https://eslint.org/docs/developer-guide/nodejs-api#linterdefineparser

uses the old "parse" API. 

Change to "parseForESLint" to reflect latest ESLint API.

Yes, I am happy to provide a PR.

cschuller:


nzakas:
Thanks for the suggestion. It's not required for parsers to use `parseForESLint()`, which is why it's not included in this page. We use `parse()` if `parseForESLint()` is not present.

PR #790, opened 2020-09-25
Update: use icons instead of text in the footer links (fixes #781)

### LINKED ISSUE #781: Using Icons insted of text-links
I guess icons will look pretty much better than text-links...
eslint

Pratyush-Saxena:


anikethsaha (changes requested):
Thanks for the PR 😃 

LGTM. 

how about adding `title` prop as well so that we get the to know what are they. it would be useful for those who are not familiar with the particular logo. 
I guess, simply using the old text as a title would help. 

Also, you need to fix the commit message and the PR title according to the format.

Pratyush-Saxena:
> Thanks for the PR 
> 
> LGTM.
> 
> how about adding `title` prop as well so that we get the to know what are they. it would be useful for those who are not familiar with the particular logo.
> I guess, simply using the old text as a title would help.
> 
> Also, you need to fix the commit message and the PR title according to the format.

pls sugggest what should i name pr title and commit message

anikethsaha:
Maybe something like this 

`Update: use icons instead of text in the footer links(fixes #781)`

nzakas:
Don’t worry about the commit message, we can fix that when we merge it. 

Adding a title attribute to the links is a good idea. :+1:

Pratyush-Saxena:
> Don’t worry about the commit message, we can fix that when we merge it.
> 
> Adding a title attribute to the links is a good idea.

@nzakas pls have a look to the changes you requested

nzakas:
Looks good to me! Thanks so much. @anikethsaha what do you think?

anikethsaha:
I was thinking of using the title attributes so that we would get the tooltip on hover effect . 
Though the current looks good to me as well.

nzakas:
I like having the text there better, personally. If you’re ok with this can you update your review?

anikethsaha (approved):
LGTM

nzakas (approved):

nzakas:
Thanks so much!

PR #804, opened 2020-10-05
New: Blog post about paying contributors

nzakas:


btmills (approved):
I really like the tone you've struck here. It's transparent about what's worked and what hasn't while phrased in a way that tries to provide useful experience for anyone reading it.

kaicataldo (approved):

PR #809, opened 2020-10-29
Fix: Update Bugsnag URL for Website

jessicaawhy:
There was no URL linked for bugsnag. This change adds in the correct URL instead of null.

### DIFF
--- a/_data/sponsors.json
+++ b/_data/sponsors.json
@@ -115,7 +115,7 @@
         },
         {
             "name": "Bugsnag Stability Monitoring",
-            "url": null,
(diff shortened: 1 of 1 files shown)


ilyavolodin:
@jessicaawhy This is an auto-generated file, it gets rebuild all the time, so this change will be lost. All you need to do to add a URL is to update your opencollective profile with the URL and auto-generation script will pick it up on the next rebuild cycle (I think it's every hour).

jessicaawhy:
Oh, I just noticed this bug and was trying to fix it. Thanks!

PR #820, opened 2020-12-23
Docs: add noreferrer to rel

chenxsan:
I believe it's very common to use `noreferrer` alongside `noopener` https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noreferrer.


nzakas:
Thanks. In this case we want to send the referrer so sponsors can see that they got traffic from us.

PR #826, opened 2021-01-31
Docs: Remove old configuring docs

btmills:
Our build step copies new files but does not delete old files, so we now have both the new `configuring/` directory and the old `configuring.md` file. The latter takes precedence in ambiguous cases. Visiting `/docs/user-guide/configuring` or `/docs/user-guide/configuring/` shows the contents of `configuring.md`. Visiting `/docs/user-guide/configuring/index.html` shows the new guide in `configuring/`.


btmills:
I've verified the new docs are working as intended in the deploy preview. Clicking "Configuring ESLint" from the "User guide" header menu or visiting https://eslint.org/docs/user-guide/configuring works as intended.

PR #830, opened 2021-02-17
Fix: Check for User and Organization sponsors (fixes #829)

### LINKED ISSUE #829: Missing Substack logo
The Substack logo is missing from the homepage. This is notable because it's one of the two sponsors we have through GitHub and it appears that the API isn't returning information for it. The other GitHub sponsor is Microsoft, but that seems to be working okay.

All of the Substack data is missing from the sponsors.json file.

nzakas:
It turns out that because Substack is an organization instead of a user, we weren't pulling that data at all. This fixes our script to pull organization data, too.

### DIFF
--- a/_tools/fetch-sponsors.js
+++ b/_tools/fetch-sponsors.js
@@ -134,10 +134,17 @@ async function fetchGitHubSponsors() {
               sponsor: sponsorEntity {
                 ...on User {
                   name,
-                 login,
-                 avatarUrl,
-                 url,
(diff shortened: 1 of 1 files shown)


mdjermanovic (approved):
Tested this locally, and it correctly adds the missing `name`, `image`, and `url` fields here:

https://github.com/eslint/website/blob/077d87f7b40f92c2758a1280fd901103098fc1d2/_data/sponsors.json#L59-L63

PR #869, opened 2021-09-17
Update: Add date posted for blog posts

thomasgilmore95:
fixes issue #833 by adding dates to blog posts.


nzakas:
Nice, simple, clean. I like it. Thank you!

nzakas (changes requested):
I appreciate all the work you put into this, but I’m afraid this isn’t scalable. Needing to manually add the date in every post isn’t practical.

a better approach would be to remove the first heading from each post (it’s duplicated in the title front matter), then modify the blog layout to insert the grading and date automatically. That will work for all posts going forward too.

thomasgilmore95:
That makes more sense to do. I actually didn't know how to do that. However, I did look it up and I will change it so the title and date are added in the blog/post layout.

thomasgilmore95:
Hi, I added the title and date to the blog post layout and removed them from the blog post. I would also add the grading, just not sure what that is.

_posts/2014-01-19-breaking-change-formatter.md
```diff
@@ -3,10 +3,9 @@ title: Breaking change to ESLint default formatter
 tags:
   - breaking change
   - formatter
+blogDate: 19 January 2014
```
nzakas, line 6: We don’t need this field because the date is in the filename. From [the Eleventy docs](https://www.11ty.dev/docs/dates/):

> If a date key is omitted from the file, we then look for a YYYY-MM-DD format anywhere in the file path (even folders). If there are multiple dates found, the first is used.

_layouts/post.liquid
```diff
@@ -12,6 +12,10 @@
       <div class="col-md-8">
         <main class="blog">
           <article>
+            <h1>{{ title }}</h1>
+            <em>{{ blogDate }}</em>
```
nzakas, line 16:
```suggestion
            <em>{{ page.date.toDateString() }}</em>
```

nzakas (changes requested):
Getting close! Just a couple bits of cleanup.

_layouts/post.liquid
thomasgilmore95: I updated to the following code in the layout: 

`<em>{{ page.date.toDateString() }}</em>`

When I run the code the following is displayed, the date is not displayed:

Screen Shot 2021-09-20 at 9 02 30 PM

_posts/2014-01-19-breaking-change-formatter.md
thomasgilmore95: I can remove the blogDate fields from the posts.

_layouts/post.liquid
mdjermanovic: I'm now sure if `liquidjs` allows function calls here. For date formatting, we're using `luxon` as a custom filter.

If we want the same date format as in the list of recent posts (which is on the same page, on the right side), it would be:

```
{{ page.date | dateFormat: "d LLLL yyyy" }}
```

An example with a formatted date is in the same file:

https://github.com/eslint/website/blob/10ecb8240da6507794d08d9937b9a0827ba01180/_layouts/post.liquid#L33

If we want another format, here's [table of tokens](https://moment.github.io/luxon/#/formatting?id=table-of-tokens).

nzakas: @thomasgilmore95 that’s weird. I took that from an Eleventy example. 

Using Luxon is a better idea anyway  👍

thomasgilmore95: Thank you. I will change to use Luxon.

_layouts/post.liquid
```diff
@@ -12,6 +12,14 @@
       <div class="col-md-8">
         <main class="blog">
           <article>
+            <h1>{{ title }}</h1>
+            {% for post in collections.posts %}
+              {% if title == post.data.title %}
+                <em>{{ post.date | dateFormat: "d LLLL yyyy" }}</em>
+              {% endif %}
+            {% endfor %}
```
mdjermanovic, line 20: Can we simplify this by taking the date from `page.date`?

```suggestion
            <em>{{ page.date | dateFormat: "d LLLL yyyy" }}</em>
```

thomasgilmore95: Updated to page.data for the date in the layout

nzakas (dismissed):
LGTM. Thanks for your patience with this!

thomasgilmore95:
@nzakas Thank you for letting me work on this.

_posts/2016-04-14-welcoming-jscs-to-eslint.md
```diff
@@ -1,4 +1,5 @@
 ---
+layout: post
```
mdjermanovic, line 2: A question - why was this added?

thomasgilmore95: I saw newer posts had that and thought I needed it for the date to show. However I found that wasn't needed. I can remove that if you would like.

mdjermanovic: You're right, newer posts have that.

It seems that in https://github.com/eslint/website/pull/649 we removed that field from all posts, but we didn't update the [template](https://github.com/eslint/eslint/blob/e3cd1414489ceda460d593ac7e7b14f8ad45d4fc/templates/blogpost.md.ejs#L2) for release blog posts, so new release posts keep producing it.

mdjermanovic: I think it's overridden at the directory level [here](https://github.com/eslint/website/blob/d44ac60c666fd1dc07ec1350bed667544af902ae/_posts/_posts.json#L2), so it might be unused at the end. Not sure how that works, though.

Either way, I can't see a difference between posts that have that field and posts that don't have it. If you also don't see a difference, then it might be better to remove it, so that we don't get confused later about why was it added in this PR.

thomasgilmore95: Removed that line from the post.

mdjermanovic (approved):
LGTM, thanks!

thomasgilmore95:
@mdjermanovic Thank you for your help.

nzakas (approved):

PR #898, opened 2021-12-24
chore: add `teaser` property to all release posts

snitin315:
add `teaser` property to all release posts.

Refers 4th point from https://github.com/eslint/website/issues/894


nzakas (changes requested):
Thanks for doing this. We also need to remove that first paragraph from each blog post body so it isn’t displayed twice:

https://deploy-preview-898--eslint.netlify.app/blog/2021/01/eslint-v7.18.0-released

nzakas (approved):
Awesome, thank you!

PR #900, opened 2021-12-31
chore: Update fetch team script to include blog authors

nzakas:
For the new website, we need a way to automatically fetch blog authors, so I updated the team fetch script to also generate an all_authors.json file. There is duplicated information between team.json and all_authors.json because we access that info in different ways in the new site.

The contributors.json file is now replaced with all_authors.json and will be autogenerated periodically through Jenkins.

Note: Once this is merged, we also need to update the Jenkins job that fetches team data to ensure that all_authors.json is committed when there are changes.


_tools/dev-server.js
```diff
@@ -107,7 +107,7 @@ app.use(morgan("tiny"));
  * TODO: Figure out why this is so that we don't have to create this custom route handler
  * and can use `express.static()` directly instead.
  */
-app.get("*", async (req, res, next) => {
+app.get("*", async(req, res, next) => {
```
nzakas, line 110: Lint fix

_tools/fetch-sponsors.js
```diff
@@ -175,7 +175,7 @@ async function fetchGitHubSponsors() {
 // Main
 //-----------------------------------------------------------------------------
 
-(async () => {
+(async() => {
```
nzakas, line 178: Lint fix

nzakas:
@eslint/eslint-tsc looks like we lost track of this. Can someone take a look?

_tools/fetch-team-data.js
```diff
@@ -71,25 +104,50 @@ const team = {
...
+    const files = (await fs.readdir(blogPostsDirectory))
+        .filter(filename => filename.endsWith(".md"))
+        .map(filename => path.join(blogPostsDirectory, filename));
+
+    for (const filename of files) {
+        const contents = await fs.readFile(filename, "utf8");
+        const { data: frontmatter } = matter(contents);
+
+        for (const username of frontmatter.authors) {
+            authors[username] = await fetchUserProfile(username);
```
btmills, line 133: Since the frontmatter is just plain strings, we should expect some usernames will have typos or just plain not exist in GitHub. What’s the desired behavior in that case? Do we blow up so we can fix it, or skip it?

btmills (commented):
Just one question about handling potentially missing usernames.

_tools/fetch-team-data.js
nzakas: Blow it up. We should never check in authors that don’t exist.

btmills: :bomb: it is!

btmills (approved):

PR #920, opened 2022-02-03
feat: Add totals to sponsors data file

nzakas:
This adds sponsorship totals to the sponsors data file. Right now we are calculating this on the homepage, but on the new website we need this data in a bunch of places, so it's easier just to have it right in the data file.


mdjermanovic (approved):
LGTM, thanks!

The numbers are slightly different now as this is data from yesterday, but the sponsor data fetch job will fix that in a few minutes so I'm merging this now.

PR #921, opened 2022-02-04
fix: Update blog post authors

### LINKED ISSUE #918: Fix blog post authors
I've been reviewing the authors listed on various blog posts, and in general, all of the posts with multiple authors listed are incorrect. It looks like people who simply contributed commits to a release are now listed as an author on the blog post, which is pretty confusing.

We need to go through and update every post with multiple authors to eliminate all but the first author that is listed. That will give us the correct info about who actually wrote the post.

For instance, if the front matter has this:

```
authors:
- nzakas
- btmills
- mdjermanovic
```

Then it should be changed to:

```
authors:
- nzakas
```

AkashaRojee:
On blog posts with more than 1 author listed, all authors except the first one have been removed.

Fixes #918

### DIFF
--- a/_posts/2014-01-19-breaking-change-formatter.md
+++ b/_posts/2014-01-19-breaking-change-formatter.md
@@ -7,7 +7,6 @@ tags:
   - formatter
 authors:
   - nzakas
-  - kaicataldo
 categories:
   - API Changes
 ---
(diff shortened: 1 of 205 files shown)


nzakas (approved):
Thank you so much for doing this!

PR #937, opened 2022-04-27
Fixed typo in code comment

AddisonG:
The example here wrongly used `x` instead of `a`.


captbaritone:
Thanks! This looks good to me.

nzakas:
Thanks! The source for this documentation is actually here:
https://github.com/eslint/eslint/blob/main/docs/src/rules/no-constant-binary-expression.md

Merging this will only affect the live site, but it will be overwritten when we do the next release. Please be sure to update the original file too.

AddisonG:
Thanks @nzakas - I already did the other one, but wasn't sure if I should do this.

nzakas (approved):

nzakas:
Ah ok. We can merge this to fix it on the live site.

